PR #6, opened 2024-11-08
CSCAP-38 Add some unit testing framework

Xerasin:
Accidently worked on CSCAP-38, wanted to get this started before my main work so I could write tests as I go. 

I put tests in backend/src/test/java/com/sim_backend as it's I believe it's preferred for jacoco. 
This just adds JUnit, Mockito, Jacoco and a Jacoco workflow to the repository.


backend/src/main/java/com/sim_backend/Main.java
```diff
@@ -33,6 +33,6 @@ public static void main(final String[] args) {
         }).start(backendPort); // Start the server
 
         // Register a test route
-        TestMessageController.registerRoutes(app);
+        //TestMessageController.registerRoutes(app);
```
pglob, line 36: Was commenting this out intentional?

backend/Dockerfile
```diff
@@ -14,7 +14,7 @@ WORKDIR /app
 RUN mvn clean package -DskipTests
 
 # Stage 2: Run the application
-FROM eclipse-temurin:21-jdk-jammy
+FROM build as run
```
pglob, line 17: Minor: I made this mistake too, but the different casing on "FROM" and "as" causes warnings. Please change "as" to "AS". Instances on lines 4, 17, and 29.

pglob:
- [ ] Make sure commits are squashed before merging.

backend/src/main/java/com/sim_backend/Main.java
Xerasin: I meant to revert this, I just needed a change in Main.java to make sure the workflow is actually picking up on the changed files.

pglob (approved):

PR #8, opened 2024-11-10
CSCAP-42/Websocket  Add Some initial Websocket code

Xerasin:
This currently all doesn't really do much, should at least function. Placing a PR now because this may help with the work for CSCAP-51 since I added some initial data structure for HeartBeat,  BootNotification and HeartBeatResponse. Also, this adds gson for serialization which may also help CSCAP-51.

Added a websocket library and a temporary client that can only send packets as of now.  I also added a JSON-schema tester so we can test the output of our messages against the OCPP-1.6 JSON Schemas.


pglob (changes requested):

backend/src/main/java/com/sim_backend/websockets/GsonUtilities.java
```diff
@@ -23,6 +24,8 @@ public static Gson getGson() {
         GsonBuilder gsonBuilder = new GsonBuilder();
```
pglob, line 24: Something I noticed in the GsonBuilder documentation:
`By default, Gson omits all fields that are null during serialization.`

The above quote may not be relevant (I don't think we're initializing fields as null ever?). However, for messages with no payload, are we serializing them with {} or null? The ocpp-j 1.6 spec requires one of them:
```Payload is a JSON object containing the results of the executed Action. If there is no payload JSON allows for two different notations: null or and empty object \{}. Although it seems trivial we consider it good practice to only use the empty object statement. Null usually represents something undefined, which is not the same as empty, and also \{} is shorter.```

File: backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java [hunk 2]
```diff
@@ -0,0 +1,189 @@
...
+    @Override
+    public void onError(Exception e) {
+
+    }
+
+    /**
+     * Set the function called when we receive an OCPP Message.
+     * @param onReceiveMessageListener The Received OCPPMessage.
+     */
+    public void setOnRecieveMessage(
```
pglob, line 140: Minor: typo, should be setOnReceiveMessage

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -0,0 +1,189 @@
...
+     * When we receive a message from a sent ocpp message.
+     * @param s The received message as a string.
+     */
+    @SuppressWarnings("checkstyle:FinalParameters")
+    @Override
+    public void onMessage(String s) {
+        if (onReceiveMessage != null) {
+            Gson gson = GsonUtilities.getGson();
+            JsonElement element = gson.fromJson(s, JsonElement.class);
+            assert element.isJsonArray();
```
pglob, line 90: We need some exception handling in case are sent a bad websocket message.

pglob (approved):
Approved provided the feedback above is addressed.

PR #14, opened 2024-11-16
Refactor JSON, tests, id generation

cfarhan:
Rewrite: 

- changed JSON array type to JSON object

- changes to HeartBeat and HeartBeatResponse 

- rewrite tests for compatibility

- generate UUIDs instead of manual id generation.

- add new Makefile command for easier testing

- reorganized code files into packages

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/messages/HeartBeat.java
+++ b/backend/src/main/java/com/sim_backend/websockets/messages/HeartBeat.java
@@ -1,9 +1,8 @@
 package com.sim_backend.websockets.messages;
 
-import com.sim_backend.websockets.OCPPMessage;
 import com.sim_backend.websockets.OCPPMessageInfo;
 
-@OCPPMessageInfo(messageName = "HeartBeat")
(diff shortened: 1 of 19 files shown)


backend/src/main/java/com/sim_backend/websockets/messages/HeartBeat.java
```diff
@@ -1,9 +1,8 @@
 package com.sim_backend.websockets.messages;
 
-import com.sim_backend.websockets.OCPPMessage;
 import com.sim_backend.websockets.OCPPMessageInfo;
 
-@OCPPMessageInfo(messageName = "HeartBeat")
+@OCPPMessageInfo(messageType = "HeartBeat", messageName = "HeartBeat")
```
cfarhan, line 5: Some confusion regarding messageType and messageName as they were being used interchangeably in some places. For now, I have both being required as part of the OCPPMessage but I think both are not required.

cfarhan:
Closing PR after reviewing the OCPP 1.6.... made the mistake of looking at 2.0.1 which is very different.

PR #16, opened 2024-11-17
Create Authorization, BootNotification, HeartBeat responses and requests

cfarhan:
Adds both the response and requests for the three message types.

Still unsure about the JSON structure for the HeartBeat request (I think an empty JSON body is what's required but I saw an empty JSON array in other documentation). 

Added Gson serialization to existing HeartBeat and BootNotification files.

Also updated OCPP message Ids to be generated via UUID's for simplicity.

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/messages/BootNotificationResponse.java
+++ b/backend/src/main/java/com/sim_backend/websockets/messages/BootNotificationResponse.java
@@ -1,0 +1,37 @@
+package com.sim_backend.websockets.messages;
+import com.google.gson.annotations.SerializedName;
+import com.sim_backend.websockets.OCPPMessage;
+import com.sim_backend.websockets.OCPPMessageInfo;
+
+@OCPPMessageInfo(messageCallID = OCPPMessage.CALL_ID_RESPONSE, messageName = "BootNotificationResponse")
+public class BootNotificationResponse extends OCPPMessage{
+
+    @SerializedName("status")
+    private String status;  // Status of the BootNotification (Accepted, Rejected)
+
+    @SerializedName("interval")
+    private int interval;   // Interval (time in seconds) for next BootNotification
+
+    // Constructor, getters, setters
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/sim_backend/websockets/messages/BootNotificationResponse.java
```diff
@@ -0,0 +1,40 @@
...
+import com.sim_backend.websockets.OCPPMessage;
+import com.sim_backend.websockets.OCPPMessageInfo;
+
+@OCPPMessageInfo(
+    messageCallID = OCPPMessage.CALL_ID_RESPONSE,
+    messageName = "BootNotificationResponse")
+public class BootNotificationResponse extends OCPPMessage {
+
+  @SerializedName("status")
+  private String status; // Status of the BootNotification (Accepted, Rejected)
```
pglob, line 13: Consider making an enum for the statuses. "Pending" can also be a status value.

backend/src/main/java/com/sim_backend/websockets/messages/BootNotificationResponse.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @SerializedName("interval")
+  private int interval; // Interval (time in seconds) for next BootNotification
+
+  // Constructor, getters, setters
+  public BootNotificationResponse(String status, int interval) {
```
pglob, line 19: Add basic validation for message fields (especially response messages). While we can catch some issues with schema validation, something like a negative interval can slip through currently.

pglob (approved):
- [ ] Create a task for next sprint to update testing with the serialization of sent messages and deserialization of received messages.

pglob:
> Still unsure about the JSON structure for the HeartBeat request (I think an empty JSON body is what's required but I saw an empty JSON array in other documentation).

I was also under the impression that empty JSON body {} (or a null) was used for messages with no payloads. Which document had the empty JSON array?

PR #17, opened 2024-11-17
Authorize   Boot Notification   Heartbeat   online/offline buttons

zhihw:
I refer to the UX design of the requirements document. In offline state, there is only a "bring online button". Once online, a drop-down menu appears with Authorize, Boot Notification, and Heartbeat buttons. They will send a request to the endpoint. The request content is currently empty. The method of sending the request includes determining whether the request is successful or rejected, as well as exception handling.

### DIFF
--- a/frontend/src/Button.js
+++ b/frontend/src/Button.js
@@ -1,7 +1,13 @@
 import React, { useState } from 'react';
 
+const messages = [
+  { name: 'Authorize', endpoint: '/api/message/authorize' },
+  { name: 'Boot', endpoint: '/api/message/boot' },
+  { name: 'Heartbeat', endpoint: '/api/message/heartbeat' },
+];
+
 //Generic method where each button POST request to the specified endpoint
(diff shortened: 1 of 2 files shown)


frontend/src/Button.js
```diff
@@ -0,0 +1,99 @@
...
+                onClick={() =>
+                  sendRequest('/api/message/authorize', 'Authorize')
+                }
+                style={{ marginBottom: '8px' }}
+              >
+                Authorize
+              </button>
+              <button
+                onClick={() =>
+                  sendRequest('/api/message/boot', 'Boot Notification')
```
pglob, line 73: Is there some way to dynamically generate these dropdown buttons from an array or list? As we add more messages, it might get difficult to read. We can rename "Boot Notification" to "Boot" if it makes mapping the name to an endpoint easier.

pglob (approved):
Looks good, just left some minor comments to address.

PR #21, opened 2024-11-19
Cscap 53 endpoints

cfarhan:
Combining this with CSCAP-51.

Implements the Authorize, BootNotification, and Heartbeat response and requests. We need to have a discussion with the frontend regarding what they need from each of these classes. They do nothing at the moment.

Also implements Auth, Boot, and Heart structures for JSON payload.

Uses MessageController class to register endpoint routes and to serve the message related endpoints.

MessageService to implement functionality of endpoints.

Adds Lombok library to help get rid of unnecessary setters and getters.

TODO:
Still need exception handling throughout code and tests for added classes.
Will also need to make a ticket to go throughout the code and make use of the Lombok class annotations to get rid of clutter code (constructors, setters, getters, etc) where possible.


Xerasin:
Why move OCPPMessage and related classes into the rest package? They are not related to the rest side of the backend at all and are only related to the websocket side.

backend/src/main/java/com/sim_backend/rest/model/BootNotification.java
```diff
@@ -0,0 +1,73 @@
+package com.sim_backend.rest.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
```
Xerasin, line 4: We are using gson for serialization.

backend/src/main/java/com/sim_backend/rest/model/BootNotification.java
```diff
@@ -0,0 +1,73 @@
...
+  private final String meterType;
+
+  /** Meter Serial Number. */
+  @SerializedName("meterSerialNumber")
+  private final String meterSerialNumber;
+
+  // Jackson annotations for deserialization
+  @JsonCreator
+  public BootNotification(
+      @JsonProperty("chargePointVendor") String chargePointVendor,
```
Xerasin, line 54: These don't take effect because we are using gson for serialization.

backend/src/main/java/com/sim_backend/rest/model/BootNotification.java
```diff
@@ -0,0 +1,73 @@
...
+  /** Meter Type. */
+  @SerializedName("meterType")
+  private final String meterType;
+
+  /** Meter Serial Number. */
+  @SerializedName("meterSerialNumber")
+  private final String meterSerialNumber;
+
+  // Jackson annotations for deserialization
+  @JsonCreator
```
Xerasin, line 52: This doesn't take effect because we are using gson for serialization.

Xerasin (changes requested):

backend/pom.xml
```diff
@@ -75,6 +75,19 @@
...
+            <artifactId>lombok</artifactId>
+            <version>RELEASE</version>
+            <scope>provided</scope>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/com.fasterxml.jackson.datatype/jackson-datatype-jsr310 -->
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>2.18.1</version>
+        </dependency>
```
Xerasin, line 89: We don't need two json libraries in our project, we already using gson for serialization.

cfarhan: Removed

backend/src/main/java/com/sim_backend/rest/model/BootNotification.java
cfarhan: Removed

pglob:
@cfarhan @Xerasin 
I'm sorry, I wasn't clear enough on the Jira tasks and didn't establish a line of communication between you guys. This resulted in conflict between your respective tasks.

Cameron, one point of clarification. The purpose behind these endpoints for the frontend to send a POST and the backend handles the request (this part is non ocpp communication). For these 3 endpoints, they will just send the 3 ocpp messages (not their responses) via the Websocket queue. 

Our simulator won't be sending/creating response messages for "Operations Initiated by Charge Point". These response messages will be received via our Websocket client and handled by a listener.

If getting the various PR's integrated with each other takes extra time, that's a ok. I can take that into consideration for next sprint's task planning for both of you.

Reach out if there is clarification I can help with.

cfarhan:
@pglob Latest commit should have the changes you are looking for. Let me know if anything else major sticks out.

cfarhan:
Going to have a new PR out for this tomorrow that cleans up a lot of the code and makes it more readable.

PR #26, opened 2024-11-27
Add simulator loop and other fixes (CSCAP-36)

pglob:
This PR is intended to allow our system to send WebSocket messages to a target server. The main addition is a simulator loop, which pops all messages in the queue every iteration. Actual integration testing for CSCAP-36 will come in a later PR.

It also refactors Main.java to be cleaner.

Below is a screenshot showing the functioning system. These messages were sent by the "Show Messages" menu in the Frontend. The server is a wscat server running in a container called "mock-server" with port 12345 exposed.

Screenshot 2024-11-26 at 8 00 58 PM


Xerasin:
Make sure you remove the dummy tests in MainTest.java, they were designed back before we ran everything in a loop so they're just causing the tests to never finish.

Xerasin (approved):

cfarhan (approved):

PR #27, opened 2024-11-28
Backend reformatting

cfarhan:
Went through some of the backend with IntelliJ and did some refactoring that it picked up.

- Like Authorize, OCPPMessage ID's generated via UUID for simplicity.
- Replace some existing setters/getters with Lombok annotations. 
- Some typos for functions and enum constants IntelliJ picked up were autocorrected. 

@JayQuack @Xerasin If removing the setters and getters isn't to your liking, we can decline the PR.

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/enums/ErrorCode.java
+++ b/backend/src/main/java/com/sim_backend/websockets/enums/ErrorCode.java
@@ -26,7 +26,7 @@ public enum ErrorCode {
    * Payload for Action is syntactically correct but at least one of the fields violates occurrence
    * constraints.
    */
-  OccurrenceConstraintViolation,
+  OccurenceConstraintViolation,
   /**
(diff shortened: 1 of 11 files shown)


backend/src/main/java/com/sim_backend/websockets/enums/ErrorCode.java
```diff
@@ -26,7 +26,7 @@ public enum ErrorCode {
    * Payload for Action is syntactically correct but at least one of the fields violates occurrence
    * constraints.
    */
-  OccurenceConstraintViolation,
```
Xerasin, line 29: The Spec has it as OccurenceConstraintViolation even though it's not correct. The code that uses this just toString this enum so make sure that code is fixed or keep it as is.

cfarhan: Reverted

pglob (approved):

PR #32, opened 2024-12-03
Charger frame easy version sample:

sille369:
** This is just a sractch UI for the charger frame.

### DIFF
--- a/frontend/src/components/charger_frame.js
+++ b/frontend/src/components/charger_frame.js
@@ -1,0 +1,82 @@
+import React from 'react';
+import { Button } from './Button';
+
+function ChargerFrame() {
+  const textContent = [
+    // { label: 'Charger Name', value: 'Sample A' },
+    { label: 'State', value: 'Available' },
+    { label: 'Meter Value', value: '123456' },
+    { label: 'Max Power', value: '40A' },
+    { label: 'Current Power', value: '0' },
+  ];
+
+  return (
+    <div
+      style={{
+        width: '300px',
+        border: '2px solid #000',
+        borderRadius: '8px',
+        padding: '20px',
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'flex-start',
+        alignItems: 'center',
+        backgroundColor: '#f9f9f9',
+      }}
+    >
+      <div
+        style={{
+          marginBottom: '20px', // Larger gap for Charger Name
+          fontSize: '18px', // Larger font for Charger Name
+          fontWeight: 'bold', // Bold Charger Name
+        }}
+      >
+        <p style={{ margin: '0', textAlign: 'left' }}>
+          <strong>Charger Name:</strong> Sample A
+        </p>
+      </div>
+
+      {/*the data displayed below for State, Meter value etc are temporary, will replace
+    with real data later*/}
+      {textContent.map((item, index) => (
+        <div key={index} style={{ marginBottom: '5px' }}>
+          {/* Smaller margin */}
+          <p style={{ margin: '0', textAlign: 'left' }}>
+            {/* Left-aligned text */}
+            <strong>{item.label}:</strong> {item.value}
+          </p>
+        </div>
+      ))}
+
+      <button
+        style={{
+          width: '200px',
+          padding: '10px',
+          margin: '10px 0',
+          border: '2px solid #000',
+          borderRadius: '8px',
+          backgroundColor: '#fff',
+          cursor: 'pointer',
+        }}
+      >
+        Plug in Vehicle
+      </button>
+      <button
+        style={{
+          width: '200px',
+          padding: '10px',
+          margin: '10px 0',
+          border: '2px solid #000',
+          borderRadius: '8px',
+          backgroundColor: '#fff',
+          cursor: 'pointer',
+        }}
+      >
+        Send custom message
+      </button>
+      <Button />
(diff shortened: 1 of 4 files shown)


zhihw:
I think the structure is good, but I'm not sure if the style parameter is applicable. We can change it later.

zhihw:
I think the structure is good, but I'm not sure if the style parameter is applicable. We can change it later.

zhihw (approved):

pglob (changes requested):
Blocking this for now, I just want to take a look before merging. Will do that soon

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Button } from './Button';
+
+function ChargerFrame() {
+  const textContent = [
+    // { label: 'Charger Name', value: 'Sample A' },
```
pglob, line 6: Remove the commented out code.

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
...
+        }}
+      >
+        <p style={{ margin: '0', textAlign: 'left' }}>
+          <strong>Charger Name:</strong> Sample A
+        </p>
+      </div>
+
+      {/*the data displayed below for State, Meter value etc are temporary, will replace
+    with real data later*/}
+      {textContent.map((item, index) => (
```
pglob, line 41: The text content should have its own component, like a "TextRow" component. They will likely be reused later, so it would be good to have.

frontend/src/components/charger_frame.js
pglob, line 1: Rename this file to ChargerFrame.js for consistency

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
...
+          {/* Smaller margin */}
+          <p style={{ margin: '0', textAlign: 'left' }}>
+            {/* Left-aligned text */}
+            <strong>{item.label}:</strong> {item.value}
+          </p>
+        </div>
+      ))}
+
+      <button
+        style={{
```
pglob, line 52: The "Plug in Vehicle" button will result in an API call when we start implementing charging sessions. This button should be built with the `ButtonBase` class in `frontend/src/components/buttons/ButtonBase.js`.

pglob (changes requested):

pglob:
> I think the structure is good, but I'm not sure if the style parameter is applicable. We can change it later.

@zhihw if you think something with the style parameter should be changed, don't be afraid to "Request changes" and comment what specifically should be changed. We have plenty of time in this project, so it is better to do it first instead of changing it later

frontend/src/components/charger_frame.js
zhihw: I think the bring online button is the outermost thing. It would be more reasonable that no button or charger status should be displayed in the offline state?

zhihw:
> > I think the structure is good, but I'm not sure if the style parameter is applicable. We can change it later.
> 
> @zhihw if you think something with the style parameter should be changed, don't be afraid to "Request changes" and comment what specifically should be changed. We have plenty of time in this project, so it is better to do it first instead of changing it later

> @pglob Just checked the current PR UI effect, I am about to modify my style to make it consistent, this style looks better on

zhihw:
> > > @pglob Just checked the current PR UI effect, I am about to modify my style to make it consistent, this style looks better on
> 
> @zhihw If the style is reused, can it be moved into a common CSS file?

@pglob I think yes, we should have at least two components, buttons and states, and the style of each component should be consistent

frontend/src/components/charger_frame.js
pglob: That's a good point. If the online/offline toggle button is in an offline state, the other buttons should be hidden. Technically, the "Plug in Vehicle" action is possible when offline, but it will be complicated to use.

I think we can still display the text information. The charger won't pause charging if it is offline; it just cannot start a new charging session.

**Let's do this: if offline, hide all buttons that result in a message being sent to the backend.**

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Button } from './Button';
+
+function ChargerFrame() {
+  const textContent = [
+    // { label: 'Charger Name', value: 'Sample A' },
+    { label: 'State', value: 'Available' },
+    { label: 'Meter Value', value: '123456' },
```
pglob, line 8: Add `Wh` for the unit of `Meter Value`. `Meter Value` represents a unit of energy, measured in watt hours.

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
...
+    { label: 'Max Power', value: '40A' },
```
pglob, line 9: Rename `Max Power` to `Max Current` for accuracy.

frontend/src/components/charger_frame.js
```diff
@@ -0,0 +1,82 @@
...
+    { label: 'Current Power', value: '0' },
```
pglob, line 10: Rename `Current Power` to `Current Flow` for accuracy.

pglob (commented):
A few more minor things.

pglob (approved):
Approved with comments

PR #34, opened 2024-12-08
CSCAP71/Add Authorization Observer

JayQuack:
Authorization Observer Added
Testing "Accepted", "Blocked" status
I think it's going to work, let me know if there is any possible improvement.


JayQuack:
I have deleted void onAuthorizeReceived(AuthorizeResponse response); from OnOCPPMessageListener interface due to error

cfarhan:
I think each individual message type needs its own MessageHandler class, or observer class, however you want to phrase it. As in, your code should be in its own "AuthorizeHandler" or "AuthorizeObserver". The current implementation will require that ALL of the individual response work must be done in onMessage found in OCPPMessageHandler, which kind of defeats the purpose of the OCPPMessageListener interface imo. @pglob @Xerasin Would like to know your insights to this. Have 

On another note I've been bogged down with other classes but will try to get my boot observer ticket out tonight. The structure of the observers might make more sense once I get a PR out.

pglob:
> I think each individual message type needs its own MessageHandler class, or observer class, however you want to phrase it. As in, your code should be in its own "AuthorizeHandler" or "AuthorizeObserver". The current implementation will require that ALL of the individual response work must be done in onMessage found in OCPPMessageHandler, which kind of defeats the purpose of the OCPPMessageListener interface imo.

Agreed, the observer needs to be in its own class. We want to avoid monoliths for processing messages as they will mostly have unique logic.

The observer can be created and registered in our main.

backend/src/main/java/com/sim_backend/websockets/observer/AuthorizeObserver.java
```diff
@@ -0,0 +1,28 @@
+package com.sim_backend.websockets.observer;
+
+import com.sim_backend.websockets.events.*;
+import com.sim_backend.websockets.exceptions.*;
+import com.sim_backend.websockets.messages.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class AuthorizeObserver {
+  private List<OCPPMessageHandler> observers = new ArrayList<>();
```
pglob, line 10: Observers should not be tracked/registered separately here. The WebSocket client has a function that allows for registering observers already (see `OCPPWebSocketClient.java`). Just register this observer in our main.

backend/src/main/java/com/sim_backend/websockets/exceptions/IllegalAuthorizeResponse.java
```diff
@@ -0,0 +1,7 @@
+package com.sim_backend.websockets.exceptions;
+
+public class IllegalAuthorizeResponse extends RuntimeException {
```
pglob, line 3: I don't think this is required. Issues with serializing / empty messages should be handled by the WebSocket client code. If I'm missing something and it is required, document the reasoning in the code like other exceptions.

pglob (changes requested):

JayQuack:
Cameron helped me to understand the concept, thanks !

backend/src/main/java/com/sim_backend/websockets/observers/AuthorizeObserver.java
pglob, line 1: Add javadoc comments

backend/src/main/java/com/sim_backend/websockets/observers/AuthorizeObserver.java
```diff
@@ -0,0 +1,42 @@
+package com.sim_backend.websockets.observers;
+
+import com.sim_backend.websockets.OCPPWebSocketClient;
+import com.sim_backend.websockets.enums.AuthorizationStatus;
+import com.sim_backend.websockets.events.*;
+import com.sim_backend.websockets.messages.*;
+
+public class AuthorizeObserver implements OnOCPPMessageListener {
+
+  public void handleAuthorize(OCPPWebSocketClient client, AuthorizeResponse response) {
```
pglob, line 10: Per the OCPP spec, the Authorize operation is only initiated by the charge point. Therefore, our simulator will never send an `AuthorizeResponse` message. 

Instead, this function can just send an `Authorize` message.

pglob (changes requested):

backend/src/test/java/com/sim_backend/websockets/observers/AuthorizeObserverTest.java
```diff
@@ -0,0 +1,44 @@
...
+  private AuthorizeObserver testObserver;
+
+  @BeforeEach
+  void setUp() {
+    MockitoAnnotations.openMocks(this);
+    testObserver = new AuthorizeObserver();
+  }
+
+  @Test
+  void testAuthorizeMessage() {
```
pglob, line 23: Capture output from `onMessageReceived` and assert that it matches the expected message. You can get a stream with something like
`private final ByteArrayOutputStream outContent = new ByteArrayOutputStream();`
 `System.setOut(new PrintStream(outContent));`
from the java.io.ByteArrayOutputStream, and java.io.PrintStream libraries.

pglob (approved):
Approved with a comment.

PR #36, opened 2024-12-10
Cscap 64 Boot Observer

cfarhan:
Creates a BootNotificationObserver:

- Has a handler to be used by the controller for requests
- Implements OnOCPPMessageListener's onMessageReceived for processing a BootNotificationResponse
- Tests created

Additionally renamed the BootNotificationStatus -> RegistrationStatus to match OCPP spec naming.

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
+++ b/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
@@ -46,7 +46,7 @@ public class MessageScheduler {
   /**
    * Our set heartbeat interval, we don't want this too common but every 4 minutes should be enough.
    */
-  private static final long HEARTBEAT_INTERVAL = 240L; // seconds
+  @Getter private static final long HEARTBEAT_INTERVAL = 240L; // seconds
 
   /** Our heartbeat job. */
   private TimedTask heartbeat;
--- a/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
+++ b/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
@@ -77,6 +77,10 @@ public class MessageScheduler {
     return (this.heartbeat = this.periodicJob(0, interval, unit, new Heartbeat()));
   }
 
+  /**
+   * Sets the OCPPTime to match that of the Central Server
+   * @param time The time of the Central Server to synchronize to
+   */
   public void synchronizeTime(ZonedDateTime time) {
     this.time.setOffset(time);
   }
--- a/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
+++ b/backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
@@ -77,6 +77,10 @@ public class MessageScheduler {
     return (this.heartbeat = this.periodicJob(0, interval, unit, new Heartbeat()));
   }
 
+  public void synchronizeTime(ZonedDateTime time) {
+    this.time.setOffset(time);
+  }
+
   /**
    * Create a Runnable instance for our job functions.
    *
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPTime.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPTime.java
@@ -67,14 +67,13 @@ public class OCPPTime implements AutoCloseable {
   }
 
   /**
-   * Synchronizes the clock to match the Central System's current time.
-   * Used when a BootNotificationResponse returns an Accepted status.
+   * Synchronizes the clock to match the Central System's current time. Used when a
+   * BootNotificationResponse returns an Accepted status.
    *
    * @param time the time you wish to sync up.
    */
-  public void setOffset(ZonedDateTime time){
-    Duration calculatedOffset =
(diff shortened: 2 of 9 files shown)


backend/src/main/java/com/sim_backend/websockets/observers/BootNotificationObserver.java
```diff
@@ -0,0 +1,74 @@
...
+    switch (response.getStatus()) {
+      case ACCEPTED:
+        // Registration successful, set heartbeat from interval
+        scheduler.setHeartbeatInterval(interval, TimeUnit.SECONDS);
+        break;
+
+      case PENDING, REJECTED:
+        // Central system is pending or rejected the request, set minimum wait time before next
+        // BootNotification request
+        if (interval <= 0) {
```
pglob, line 62: If the interval is negative, that should result in our simulator responding with a CallError. Since we haven't implemented that yet, I think an exception is appropriate? That way it will trace back to the WebSocket client.

backend/src/main/java/com/sim_backend/websockets/observers/BootNotificationObserver.java
```diff
@@ -0,0 +1,74 @@
...
+import com.sim_backend.websockets.events.OnOCPPMessageListener;
+import com.sim_backend.websockets.messages.BootNotification;
+import com.sim_backend.websockets.messages.BootNotificationResponse;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Observer that handles BootNotification requests and responses as part of the OCPP protocol
+ * communication with the Central System.
+ */
+public class BootNotificationObserver implements OnOCPPMessageListener {
```
pglob, line 18: The BootNotificationObserver also needs to subscribe to state changes. When we move into a BootingUp state, it should automatically send the BootNotification message.

backend/src/main/java/com/sim_backend/websockets/observers/BootNotificationObserver.java
```diff
@@ -0,0 +1,74 @@
...
+
+    if (!(message.getMessage() instanceof BootNotificationResponse response))
+      throw new ClassCastException("Message is not a BootNotificationResponse");
+    MessageScheduler scheduler = message.getClient().getScheduler();
+    long interval = response.getInterval();
+
+    switch (response.getStatus()) {
+      case ACCEPTED:
+        // Registration successful, set heartbeat from interval
+        scheduler.setHeartbeatInterval(interval, TimeUnit.SECONDS);
```
pglob, line 56: >... it is RECOMMENDED to synchronize its internal clock with the supplied Central System’s current time. 

Given this, we should also sync the clock here. I think we need to add functionality in OCPPTime.java to set `offset` outside of a Heartbeat message.

backend/src/main/java/com/sim_backend/websockets/observers/BootNotificationObserver.java
```diff
@@ -0,0 +1,74 @@
...
+        break;
```
pglob, line 57: I had forgotten this in the task description, but add a state transition to `Available` if the BootNotification is accepted.

pglob (changes requested):

pglob:
One more thing, create and hook the observer in our main. Unless I'm missing something, we can have the BootNotification operation working then.

cfarhan:
If @Xerasin could check the Synchonize commit and give any feedback would be great. Probably a more intuitive way of synchronizing the time.

backend/src/main/java/com/sim_backend/websockets/MessageScheduler.java
```diff
@@ -77,6 +77,10 @@ public TimedTask setHeartbeatInterval(Long interval, TimeUnit unit) {
     return (this.heartbeat = this.periodicJob(0, interval, unit, new Heartbeat()));
   }
 
+  public void synchronizeTime(ZonedDateTime time) {
```
pglob, line 80: Add a javadoc comment here

pglob (approved):
Approved with minor comments

cfarhan:
Will see if Katie has any objections to the time synchronize I implemented before merging.

backend/src/main/java/com/sim_backend/websockets/OCPPTime.java
```diff
@@ -66,6 +66,17 @@ public ZonedDateTime getSynchronizedTime(ZonedDateTime time) {
     return time.plus(offset.get());
   }
 
+  /**
+   * Synchronizes the clock to match the Central System's current time. Used when a
+   * BootNotificationResponse returns an Accepted status.
+   *
+   * @param time the time you wish to sync up.
+   */
+  public void setOffset(ZonedDateTime time) {
```
Xerasin, line 75: The above listener uses the exact same code, make it use this method.

```java
  @VisibleForTesting
  final OnOCPPMessageListener listener =
      message -> {
        HeartbeatResponse response = (HeartbeatResponse) message.getMessage();

        this.setOffset(response.getCurrentTime());
      };
```

Xerasin (changes requested):

Xerasin (approved):

PR #38, opened 2025-01-10
Run frontend tests on PRs

pglob:
Also change backend coverage runs to only run when a file in backend/** was changed.


pglob:
Just testing out stuff

PR #40, opened 2025-01-12
postrequest can handle text and add some test

zhihw:


frontend/src/components/buttons/ButtonBase.js
```diff
@@ -10,7 +10,7 @@ class ButtonBase {
     // TODO: Look into options for dynamically setting this url.
     // Something like `${window.location.protocol}//${window.location.hostname}:8080`.
     // The above works for external browsers, but not the internal docker network.
-    const url = `${process.env.REACT_APP_BACKEND_URL}${this.endpoint}`;
+    const url = `http://localhost:8080${this.endpoint}`;
```
pglob, line 13: Is there a reason this was changed? Making this `localhost:8080` breaks the internal network used during integration testing.

package.json
pglob, line 1: What's the reason for this file being outside of `frontend/**`?

package-lock.json
pglob, line 1: What's the reason for this file being outside of `frontend/**`?

pglob (commented):
Just a few questions, looks good otherwise.

package.json
zhihw: Sorry they shouldn't be here, I was just trying to install some tools while writing code I'll re-PR

PR #44, opened 2025-01-19
Status Notification

zhihw:
There are two problems. The first is that the ocpp protocol requires two enumeration classes for the status class variables. I am not sure what should be enumerated in them, such as which error codes we need. In addition, I am not sure whether the constructor needs to create some values ​​in the /constant like boot.

### DIFF
--- a/backend/src/main/java/com/sim_backend/rest/controllers/MessageController.java
+++ b/backend/src/main/java/com/sim_backend/rest/controllers/MessageController.java
@@ -13,6 +13,11 @@ import com.sim_backend.websockets.messages.Authorize;
 import com.sim_backend.websockets.messages.BootNotification;
 import com.sim_backend.websockets.messages.Heartbeat;
 import com.sim_backend.websockets.messages.StatusNotification;
+import com.sim_backend.websockets.enums.ChargePointErrorCode;
+import com.sim_backend.websockets.enums.ChargePointStatus;
+import java.time.OffsetDateTime;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import io.javalin.Javalin;
 import io.javalin.http.Context;
 import java.net.URI;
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/sim_backend/rest/controllers/MessageController.java
```diff
@@ -69,12 +70,19 @@ public void offline(Context ctx) {
     ctx.result("OK");
   }
 
+  public void status(Context ctx) {
```
pglob, line 73: This endpoint will be different than other endpoints. To allow the frontend to send a specific error, this endpoint needs to parse a message. Determine which format you want to use.

Possible example:
The frontend sends 
```
{
  "connectorId": "1",
  "errorCode": "HighTemperature",
  "info": "",
  "status": "Faulted",
  "timestamp": "",
  "vendorId": "",
  "vendorErrorCode": ""
}
```
The endpoint function parses that message and sends the nonempty arguments to the StatusNotification constructor.

This way, you can have an error menu like you showed in our UI meeting. That error menu could then result in specific StatusNotification messages being sent.

backend/src/main/java/com/sim_backend/websockets/messages/StatusNotification.java
```diff
@@ -0,0 +1,55 @@
...
+  private OffsetDateTime timestamp;
+
+  @SerializedName("vendorId")
+  private String vendorId;
+
+  @SerializedName("vendorErrorCode")
+  private String vendorErrorCode;
+
+  /*Not sure how to structure it yet,use default values now*/
+  public StatusNotification() {
```
pglob, line 46: No default values are required here, the constructor should expect all required fields to be passed in. Optional fields can default to empty.

pglob (changes requested):

PR #46, opened 2025-01-19
Cscap 102/stopping charge

JayQuack:
Stopping Charge Handler, Request, Response Added


backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+    this.client = client;
+  }
+
+  /**
+   *  Authorization Process before initiating Stop Transaction
+   *  
+   * @param idTag ID of an user
+   * @throws IllegalStateException If stateMachine status is not a Charging state throw error
+   */
+  public void preAuthorize(String idTag) {
```
pglob, line 35: `preAuthorize` is almost duplicated between the Stop and Start transaction handlers. Consider utilizing the `AuthorizeObserver` in these files (with some changes to make it possible to retrieve the auth status)

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+    // transition to Available, check if StateMachine is Charging
+    if (stateMachine.getCurrentState() != SimulatorState.Charging) {
+      throw new IllegalStateException(
+          "Cannot stop with the current state. current state: " + stateMachine.getCurrentState());
+    }
+
+    Authorize authorizeMessage = new Authorize(idTag);
+    client.pushMessage(authorizeMessage);
+
+    client.onReceiveMessage(
```
pglob, line 45: In both the start and stop transaction files, I believe we need to remove the listener once we're done with it. Otherwise they will stack up. Let me know if my understanding is incorrect.

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
pglob, line 1: This file and `StartTransactionHandler.java` need some logic to make starting and stopping transactions a single function call.

Changes for `StartTransactionHandler.java`: 
>`stateMachine.transition(SimulatorState.Preparing);` on line 55 needs to occur before authorization, and then transition back into `Available` if the idTag was rejected.

> Also create a function that runs `preAuthorize()` and then calls `initiateStartTransaction()` depending on its result.

Changes for `StopTransactionHandler.java`:
>We need a single function that calls `preAuthorize()` if the idTag DOES NOT match the idTag used for starting the transaction. If the authorize was successful or the idTag DOES match the idTag used for starting the transaction, stop the transaction.

pglob (changes requested):
To have access to the StartTransactionHandler, you can rebase this branch onto main.

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
JayQuack: I added clearonreceivedmessage, how about authorizeresponse ?

JayQuack:
I tried to use Ahtorizationobserver and it became a spaghetti code. I tried to make it simple as possible

pglob:
Could you clean up your branch? It's hard to review with the duplicated file changes.

PR #51, opened 2025-01-20
Create config registry

cfarhan:
Created a configuration registry for a simulator instance.

Couldn't really think of much else to add here. Let me know if you think the BootNotification variables (like chargePointModel) are not needed, I'm leaning on not including them but let me know.

### DIFF
--- a/backend/src/main/java/com/sim_backend/ConfigurationRegistry.java
+++ b/backend/src/main/java/com/sim_backend/ConfigurationRegistry.java
@@ -1,7 +1,7 @@
 package com.sim_backend;
 
-import com.sim_backend.state.SimulatorState;
 import com.sim_backend.websockets.constants.BootNotificationConstants;
+import com.sim_backend.websockets.enums.MeterValuesSampledData;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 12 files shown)


backend/src/main/java/com/sim_backend/ConfigurationRegistry.java
pglob, line 1: From an email Brandon sent to us:
```
Following up from our call today, here are the key/values that we want to support with the ChangeConfiguration request

MeterValuesSampledData (String): Current.Offered,Current.Import,Energy.Active.Import.Register,Energy.Active.Import.Interval,Power.Active.Import,Power.Offered
MeterValueSampleInterval (integer, "30")
AuthEnabled (boolean, true)
```

We need to be able to store these values as well. MeterValueSampleInterval would have a default value of 30. AuthEnabled would default to true. MeterValuesSampledData defaults to "Energy.Active.Import.Register"

backend/src/main/java/com/sim_backend/ConfigurationRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+  private String authKeyValue;
+
+  /** Authorize response id tag */
+  private String idTag;
+
+  /** The ChargeBox's Serial Number */
+  private String centralSystemUrl;
+
+  /** Current state of the simulator */
+  private SimulatorState simulatorState;
```
pglob, line 31: This one could be removed. It feels like less of a configuration since it can change without user interaction.

pglob (approved):
Just left a couple comments. I would lean into keeping BootNotification variables in this. That would make it easier for the frontend to modify those values.

PR #53, opened 2025-01-21
Retrieve logged message in the backend

sille369:
* Created API endpoints on the backend for retrieving Tx/Rx message
* Frontend can retrieve these messages
* Created initial UI for log messages

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
@@ -93,7 +93,9 @@ public class OCPPWebSocketClient extends WebSocketClient {
    * @param message The received message.
    */
   public void recordRxMessage(String message) {
-    rxMessages.add(message);
+    String timestamp = OffsetDateTime.now(ZoneOffset.UTC).toString();
+    String messageWithTimestamp = message.replaceFirst("\\[", "[\"" + timestamp + "\", ");
+    rxMessages.add(messageWithTimestamp);
   }
 
   /**
@@ -213,7 +215,6 @@ public class OCPPWebSocketClient extends WebSocketClient {
     }
 
     OCPPMessage message = (OCPPMessage) gson.fromJson(data, messageClass);
-    rxMessages.add(s); // Store the received message
     message.setMessageID(msgId);
     this.handleReceivedMessage(messageClass, message);
   }
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
@@ -314,11 +314,10 @@ public class OCPPWebSocketClient extends WebSocketClient {
    *
    * @param message the message to be sent.
    */
-  public void pushMessage(final OCPPMessage message) {
-    queue.pushMessage(message);
+  public boolean pushMessage(final OCPPMessage message) {
     recordTxMessage(message.toJsonString()); // Record transmitted message
+    return queue.pushMessage(message);
   }
-
   /**
    * Return the size of the send queue.
    *
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
@@ -27,6 +27,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.drafts.Draft_6455;
 import org.java_websocket.handshake.ServerHandshake;
+import java.util.List;
 
 /** A WebSocket client for handling OCPP Messages. */
 @Slf4j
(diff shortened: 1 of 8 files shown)


frontend/src/components/ShowLogMessage.js
pglob, line 1: Some general thoughts on ways of making the display more concise:
* The `Time:`, `Message Type:` and `Number ID:` labels are not necessary.
* Along those lines, the date part of the time stamp is not necessary. It would not be required information for the user in almost all cases.
* Instead of showing the number id as a numeric value, it would be great if it would be converted into `Call` `Response` or `Error`.
* By doing the above, we could probably fit the initial information into 1 line.  A line would then look like:
`6:16:18 PM - BootNotification Call`
* With the `Show Details` button, can that functionality be moved onto the line of text itself? Aka, if the `6:16:18 PM - BootNotification Call` text is clicked, it will show the contents. If multiple messages are in the log, this would save a lot of space and the user would not see `Show Details` over and over. Having some color feedback when mousing over the text would indicate that it is clickable.

pglob (changes requested):

frontend/src/components/ShowLogMessage.js
sille369: I made some new updates in terms of the UI of the log messages. And nice idea of making the message into one line, will do it :)

sille369: Updated ! There is a new look for the UI of the log message now :)

pglob (commented):
Oops, I had a review typed out a few days ago but forgot to submit it.

The color idea for "Call", "Result" and "Error" is good. I would rename "Result" to "Response". I feel it's more clear outside of the OCPP-J document. It's not OCPP terminology, so push back if you feel that it should be "Result".

To put emphasis on the message type (like `Heartbeat`, `Authorize` etc), consider making the timestamp not bold.

Let's go ahead and make the log less wide like Cameron suggested. Line it up with the charger frame width. You can remove the date part of the time stamp to make it fit. A log entry would just look like `4:37:46 PM - Heartbeat Call` then.

Regarding the received messages, there seems to be an issue parsing them. I see them as `1/28/2025, 4:35:30 PM - [object Object] Result` and there is a parsing error when clicking on it. If you need to do backend debugging, we have a `make docker_debug` option. In vscode, you can use our `launch.json` to attach the debugger.

sille369:
I had no idea why Measure coverage action fail.

pglob:
> I had no idea why Measure coverage action failed.

I have an idea of why, I'll take a further look when this gets closer to merging.

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -67,6 +70,54 @@ public class OCPPWebSocketClient extends WebSocketClient {
...
+
+  /**
+   * Record a received message.
+   *
+   * @param message The received message.
+   */
+  public void recordRxMessage(String message) {
+    String timestamp = OffsetDateTime.now(ZoneOffset.UTC).toString();
+    // Replace the currentTime value using regex
+    String modifiedMessage =
```
pglob, line 98: Why are you modifying the Rx message instead of appending the timestamp like you did above? `currentTime` is an important field in some messages.

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -96,6 +147,7 @@ public void onOpen(ServerHandshake serverHandshake) {
   public void onMessage(String s) {
     try {
       this.handleMessage(s);
+      this.recordRxMessage(s); // Record received message
```
pglob, line 150: Looking at the Rx messages in the UI, responses don't show the message name. `this.recordRxMessage(s)` will need to be moved into the individual cases inside `handleMessage()`. By doing this, you can get the message name for response and error messages with the `OCPPMessage prevMessage = this.queue.getPreviousMessage(msgId);` that's on line 184 and 196. `prevMessage.messageName()` should get you the message name you need.

This way our UI for response messages can show what it is responding to. Like `2:29:24 PM - Heartbeat Response`

frontend/src/components/ShowLogMessage.js
```diff
@@ -0,0 +1,237 @@
...
+
+  // State for dropdown menu selection (log message type)
+  const [logMessageType, setLogMessageType] = useState('default');
+
+  // Helper function for mapping NumId to labels
+  const getNumIdLabel = (NumId) => {
+    return (
+      {
+        2: 'Call',
+        3: 'Respone',
```
pglob, line 20: Typo, should be `Response`

pglob (changes requested):
Just a few more things. 

Can we revert the design back to when there was no drop down? So both the Tx and Rx messages are visible at the same time? 

From a user perspective, they likely care about a whole "interaction". Being able to see a sent message and the corresponding received message without navigating a dropdown has value. If you disagree, let me know and we can discuss. (If the reason for the dropdown was the increased size of the ChargerFrame, I am shrinking it back to the original size in a future PR)


Due to the complexity here, add some frontend unit testing for the parsing.

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
sille369: I wonder, so the currentTime is not the same as the timestamp when we response the message back?

pglob: `currentTime` might be the same as timestamp, but it is not guaranteed. `currentTime` is based off the Central System clock. Regardless, the purpose of our simulator is testing OCPP Central Systems, therefore, we cannot change message contents. If we change the contents, testing that Ranger EV does may be invalid.

pglob:
> I had no idea why Measure coverage action failed.

@sille369 Replace all instance of `OffsetDateTime` in our codebase to `ZonedDateTime`. The issue is `recordTxMessage(message.toJsonString());` immediately serializes the message with GSON. We do not have a custom serializer for `OffsetDateTime`, only `ZonedDateTime`. This did not fail before because the StatusNotification message never got sent by the message queue, so it was never serialized.

frontend/src/components/ShowLogMessage.js
```diff
@@ -0,0 +1,248 @@
...
+            </p>
+
+            {/* Message Number ID (numIdLabel) with a rectangular box */}
+            <div className={`num-id-label ${numIdStyle}`}>{numIdLabel}</div>
+          </div>
+          {/* Expanded Details */}
+          {expandedSentMessages.has(userId) && (
+            <div>
+              <p>
+                <strong>User ID:</strong> {userId}
```
pglob, line 103: Rename this to something like `Unique ID` or `Message ID`. `User ID` might imply a false relation to idTags.

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -67,6 +70,53 @@ public class OCPPWebSocketClient extends WebSocketClient {
...
+    rxMessages.add(modifiedMessage);
+  }
+
+  /**
+   * Get the list of transmitted messages.
+   *
+   * @return List of transmitted messages.
+   */
+  public List<String> getSentMessages() {
+    return txMessages;
```
pglob, line 108: To keep the number of messages to a reasonable length, consider having `getSentMessages()` and `getReceivedMessages()` only return the most recent 20 messages or so. (Leaving the simulator idle could create a TON of heartbeats after a while)

pglob (approved):
Address these and you're good to go. Make sure to follow the comment I left above about the `OffsetDateTime` and `ZonedDateTime` to fix the failing unit test

PR #55, opened 2025-01-26
common styles

zhihw:
Use styles.css to render the component. In addition, make the charger frame longer. Now the length is just enough to ensure that the opened dropdown menu is in the frame.


pglob (approved):
Looks good, just one thing. Could you change the charger frame back to its original size? That extra space underneath would get used by the message logs.

I'll make a task for figuring out to the best way to approach the dropdown clipping. We might discuss it in a future UI meeting

PR #56, opened 2025-01-28
Electrical Interface

cfarhan:
Creates an electrical interface that currently just manages electrical values when changing state.
All values are initialized to 0 since we begin not charging and are updated once charging begins.

One question I had was if both getEnergy() methods needs to take into account the charging not being fully active from the initial timestamp. Maybe this will be implemented later. For now, it assumes that once charging starts, it remains charging with no interruptions.

### DIFF
--- a/backend/src/main/java/com/sim_backend/ElectricalTransition.java
+++ b/backend/src/main/java/com/sim_backend/ElectricalTransition.java
@@ -26,10 +26,13 @@ public class ElectricalTransition {
   private long initialChargeTimestamp;
 
   /** Constant representing the number of seconds in an hour. Used for energy calculations. */
-  private static final float SECONDS_PER_HOUR = 3600.0f;
+  private static final long SECONDS_PER_HOUR = 3600;
 
   /** Constant representing the number of milliseconds in an hour. Used for energy calculations. */
   private static final long MILLISECONDS_PER_HOUR = 3600000;
+
+  /** Constant representing the number of milliseconds in a second. Used for energy calculations. */
+  private static final long MILLISECONDS_PER_SECOND = 1000;
 
   /**
    * Calculates and returns the maximum power the charger is capable of providing in kilowatts (kW).
@@ -56,6 +59,12 @@ public class ElectricalTransition {
    * @return the energy consumed since the given interval in kilowatt-hours (kWh).
    */
   public float getEnergyActiveImportInterval(int interval) {
+    long timeChargingSeconds = (System.currentTimeMillis() - this.initialChargeTimestamp) / MILLISECONDS_PER_SECOND;
+
+    // If the requested interval exceeds the actual charging time, adjust accordingly
+    if(timeChargingSeconds < interval) {
+      interval = (int) timeChargingSeconds;
+    }
     return getPowerActiveImport() * ((float) interval / SECONDS_PER_HOUR);
   }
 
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/sim_backend/ElectricalTransition.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * Returns the lifetime energy used by the charger in kilowatt-hours (kWh).
+   *
+   * @return the lifetime energy consumed in kilowatt-hours (kWh).
+   */
+  public float getEnergyActiveImportRegister() {
+    long timeCharging = System.currentTimeMillis() - this.initialChargeTimestamp;
+    return this.powerOffered * ((float) timeCharging / MILLISECONDS_PER_HOUR);
```
pglob, line 60: Use `powerActiveImport`. While `powerActiveImport == powerOffered` for now, when load balancing is added, that may not be true.

backend/src/main/java/com/sim_backend/ElectricalTransition.java
```diff
@@ -0,0 +1,91 @@
...
+  private static final long MILLISECONDS_PER_HOUR = 3600000;
+
+  /**
+   * Retrieves the amount of energy consumed since the specified interval in kilowatt-hours (kWh).
+   *
+   * @param interval the interval in seconds for which the energy usage is being queried
+   * @return the energy consumed since the given interval in kilowatt-hours (kWh).
+   */
+  public float getEnergyActiveImportInterval(int interval) {
+    return this.powerOffered * ((float) interval / SECONDS_PER_HOUR);
```
pglob, line 50: Use `powerActiveImport`. While `powerActiveImport == powerOffered` for now, when load balancing is added, that may not be true.

backend/src/main/java/com/sim_backend/ElectricalTransition.java
pglob, line 1: Add getters for the rest of the electrical parameters here. The frontend can then display them and we will be sending them via `MeterValue` messages in the future.

backend/src/main/java/com/sim_backend/ElectricalTransition.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Initiates the charging transition, setting the initial charging parameters such as voltage,
+   * current, and power. The function also checks if the state machine is in the correct state
+   * (Charging) before applying the transition.
+   *
+   * @param stateMachine The state machine controlling the charging process.
+   * @throws IllegalStateException If the current state is not 'Charging', an exception is thrown.
+   */
+  public void ChargingTransition(SimulatorStateMachine stateMachine) {
```
pglob, line 71: Sorry I should have put this in the task description. Instead of implementing this as a transitory structure, this should be a permanent component in our charger. If the state isn't `Charging`: `voltage`, `currentOffered` and `currentImport` would be 0. The remaining electrical values would be 0 as a result.

Having the object available at all times simplifies the interactions with it.

pglob (changes requested):
For the `getEnergyActiveImportInterval()` function, take into account the initial timestamp. This energy will increase once charging starts, until after `interval` seconds, where it will become stable. 

Both getEnergy() functions can assume that the current/power is stable for now. This will no longer be true once load balancing is implemented. Load balancing can limit the power (thus the current). This would require actively tracking the energy over time with a summation. You would likely get the task for adding active tracking to this component, so I leave it up to you if you want to implement it now.

cfarhan:
Will create a test for it either tomorrow or Wednesday.

backend/src/main/java/com/sim_backend/ElectricalTransition.java
pglob, line 1: One more thing, implement the `StateObserver` to get automatically notified of state changes.

pglob (approved):

PR #57, opened 2025-01-30
Offline Features

Xerasin:
Added the ability for our client to go "offline." I also added a queue for messages we received when offline, since websockets are TCP, this was meant to simulate us receiving a ton of packets once the machine gets back online. I'm unsure if this is the right way to handle this though.

Stopping the ping-pong is done via stopConnectionLostTimer.


pglob:
I think I'll have to send Ranger EV an email for clarification on this behavior.

If the offline period is short enough, the TCP retries should inherently recreate "us receiving a ton of packets once the machine gets back online". With the implementation you have, the connection would still be active from the TCP perspective. 

I'll ask Ranger EV which of the below they want
> Options for offline behavior:
> * Ignoring all OCPP messages we receive and stopping WebSocket ping pong. TCP ACKS would still get sent
> * Closing the socket directly, but I believe this may send a TCP RST
> * Since the backend can deploy in a docker container, it would probably be necessary to drop nonfrontend packets from the network interface

Are there any other points you'd want included in the email?

Xerasin:
That sounds good to me! Yeah I wasn't quite sure what they were wanting here

cfarhan (approved):

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -56,6 +58,12 @@ public class OCPPWebSocketClient extends WebSocketClient {
   /** The OCPP Message Queue. */
   private final MessageQueue queue = new MessageQueue();
 
+  /** Our received messages while offline */
+  private final Deque<String> offlineQueue = new LinkedList<>();
```
pglob, line 62: I saw we remove the `offlineQueue`. This would give Ranger EV a better way to test their system (it forces them to retry messages instead of just waiting until the offline button is clicked again).

pglob (approved):
@Xerasin got an email back from Ranger EV:
> For taking the charger offline, this is just so we can simulate the backend not being able to connect to the charger (unresponsive for some reason). This will allow us to test features like monitoring the charger status and sending out emails (e.g. "connection lost", "connection reestablished" etc). In the real world, this happens when there is either a network issue or power outage. We saw this recently when power went out in Manzanita this winter (where our test deployments are). From the backends perspective, any of the "other options" would probably work. I don't think we are listening to anything on the TCP level for understanding charger status yet. ... [Brandon] if you want to opine here at all jump in.

I say we stick with the current implementation. Please test the current implementation with another WebSocket server and something like Wireshark. I'd like the behavior of this offline mode characterized. It's ok if other tasks get punted into next sprint.

I'll check back in with Ranger EV when its merged.

PR #60, opened 2025-01-30
Start, Stop Charging Handler

JayQuack:
Start Stop Charging Handler
Transaction Handler added

Sorry I made a branch a lot worse than ever before.
I created a new one


JayQuack:
Next commit will pass spotless check

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    Authorize authorizeMessage = new Authorize(idTag);
+    client.pushMessage(authorizeMessage);
+
+    client.onReceiveMessage(
+        AuthorizeResponse.class,
+        message -> {
+          if (!(message.getMessage() instanceof AuthorizeResponse response)) {
+            throw new ClassCastException("Message is not an AuthorizeResponse");
+          }
+          stateMachine.transition(SimulatorState.Preparing);
```
pglob, line 51: In this case we don't transition to `Preparing` since a transaction is currently active.

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+        StopTransactionResponse.class,
+        message2 -> {
+          if (!(message2.getMessage() instanceof StopTransactionResponse response2)) {
+            throw new ClassCastException("Message is not a StopTransactionResponse");
+          }
+          if (response2.getIdTaginfo().getStatus() == AuthorizationStatus.ACCEPTED) {
+            System.out.println("Transaction Completed...");
+            stateMachine.transition(SimulatorState.Available);
+          } else {
+            System.err.println("Transaction Failed...");
```
pglob, line 93: Put a bit more detail in most of these logs. `Transaction Failed...` implies that the transaction might have not started. This really should be something like `Transaction failed to stop`

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+            stateMachine.transition(SimulatorState.Charging);
+          }
+        });
+
+    client.clearOnReceiveMessage(StopTransactionResponse.class);
+  }
+
+  public void StopCharging(String preAuthidTag, String postAuthidTag, int transactionId) {
```
pglob, line 101: Add javadoc

backend/src/main/java/com/sim_backend/transactions/StartTransactionHandler.java
```diff
@@ -102,5 +102,17 @@ public void initiateStartTransaction(int connectorId, String idTag) {
             stateMachine.transition(SimulatorState.Available);
           }
         });
+
+    client.clearOnReceiveMessage(StartTransactionResponse.class);
+  }
+
+  /** Authorize user and initiate charging */
+  public void preAuthStartCharging(int connectorId, String idTag) {
```
pglob, line 110: Add javadoc

pglob, line 110: Name this function `StartCharging` so it is consistent with `StopCharging`

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    if (preAuthidTag != postAuthidTag) {
+      preAuthorize(postAuthidTag);
```
pglob, line 103: The result of this authorize needs to be used

backend/src/main/java/com/sim_backend/websockets/messages/StartTransaction.java
```diff
@@ -13,7 +12,6 @@
  */
 @Getter
 @Setter
-@EqualsAndHashCode(callSuper = true)
```
pglob, line 16: Is there a reason for removing this?

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,47 @@
...
+  private StopTransactionHandler stopHandler;
+
+  // Constructor
+  public TransactionHandler(SimulatorStateMachine stateMachine, OCPPWebSocketClient client) {
+    startHandler = new StartTransactionHandler(stateMachine, client);
+    stopHandler = new StopTransactionHandler(stateMachine, client);
+  }
+
+  // Initial Start Charging for the first time
+  public void PreAuthlStartCharging(int connectorId, String idTag) {
```
pglob, line 24: `PreAuthlStartCharging` and `PostAuthStartCharging` should be 1 function.

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+          }
+        });
+  }
+
+  /**
+   * Initiate Stop Transaction Handling StopTransaction Request, Response and simulator status If
+   * authorization is accepted, change a stateMachine status from Charging to Available Stays
+   * Charging otherwise
+   */
+  public void initiateStopTransaction(int transactionId) {
```
pglob, line 68: The idTag, and reason need to be optional fields in the stopTransactionMessage

pglob (changes requested):
To achieve the flow described in `3.4. General views of operation` I propose you do the following:

- Have the lambda's in the `preAuthorize` functions call `initiateStartTransaction` and `initiateStopTransaction`. That way, we can chain the functions together and it does not require external checks to run. `initiateStopTransaction` should still be called directly if the `preAuthorize` is not required.
- It is required that the transactionId be saved and stored in `TransactionHandler`. `TransactionHandler` can then pass that to the `StopCharging` function.
- For simplicity, ignore the `ConfigurationRegistry`. `TransactionHandler` needs to save the `idTag` passed in to `StartCharging`.
- Remove the Authorize observer files. With the implementation you've done here, we won't be using it.

The idea behind this is that someone can call `StartCharging` or `StopCharging` in the `TransactionHandler` and it will do the work in one step.
Let me know if you have questions, we can also have a meeting going over it.

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+  private OCPPWebSocketClient client;
+
+  // Constructor
+  public StopTransactionHandler(SimulatorStateMachine stateMachine, OCPPWebSocketClient client) {
+    this.stateMachine = stateMachine;
+    this.client = client;
+  }
+
+  /** Initiate Stop Transaction Handling StopTransaction Request, Response and simulator status */
+  public void initiateStopTransaction(int transactionId, String idTag) {
```
pglob, line 29: Add parameters to the javadoc.

Some grammar fixes: 
- Make `Stop Transaction` 1 word, `StopTransaction`
- Add a comma before `Handling`
- Change `Handling` to `handles`
- Make `Request` and `Response`, `requests` and `responses`

backend/src/main/java/com/sim_backend/transactions/StopTransactionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+import com.sim_backend.websockets.OCPPTime;
+import com.sim_backend.websockets.OCPPWebSocketClient;
+import com.sim_backend.websockets.enums.*;
+import com.sim_backend.websockets.messages.*;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import lombok.Getter;
+
+/*
+ * This class handles sending an authorize request, followed by a StopTransaction message.
```
pglob, line 14: This class no longer handles auth requests, update the comment

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+
+import com.sim_backend.state.SimulatorState;
+import com.sim_backend.state.SimulatorStateMachine;
+import com.sim_backend.websockets.OCPPWebSocketClient;
+import com.sim_backend.websockets.enums.AuthorizationStatus;
+import com.sim_backend.websockets.messages.Authorize;
+import com.sim_backend.websockets.messages.AuthorizeResponse;
+import lombok.Getter;
+
+/** This TransactionHandler manages Start and Stop opreation */
```
pglob, line 11: Typo: `opreation` should be `operation`

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    startHandler = new StartTransactionHandler(stateMachine, client);
+    stopHandler = new StopTransactionHandler(stateMachine, client);
+    this.stateMachine = stateMachine;
+    this.client = client;
+    this.idTag = null;
+    this.transactionId = -1;
+  }
+
+  /**
+   * Authorization Process before Start Transaction When Authorize is accepted, change a
```
pglob, line 32: Make sure to mention this functions use for `StopTransaction` as well. 

This function doesn't directly make state changes, so that can be dropped. Instead describe that it starts or stops a transaction depending on the current state and authorization result

Grammar fixes:
-  Change `Process` to `process`
- Make `Start Transaction` 1 word, `StartTransaction`
- Add a period before `When`

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+          }
+
+          if (response.getIdTagInfo().getStatus() == AuthorizationStatus.ACCEPTED) {
+            System.out.println("Authorization Accepted...");
+            System.out.println("Proceeding Transaction...");
+            if (stateMachine.getCurrentState() == SimulatorState.Preparing) {
+              startHandler.initiateStartTransaction(connectorId, idTag);
+              this.transactionId = startHandler.getTransactionId();
+              this.idTag = idTag;
+              stateMachine.transition(startHandler.getStateMachine().getCurrentState());
```
pglob, line 56: The `stateMachine` instance inside the start and stop handlers is the same instance as the `stateMachine` in `TransactionHandler`. Therefore, transitions made inside the handlers reflect in `TransactionHandler`. The extra transitions are redundant and can be removed.

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    client.onReceiveMessage(
+        AuthorizeResponse.class,
+        message -> {
+          if (!(message.getMessage() instanceof AuthorizeResponse response)) {
+            throw new ClassCastException("Message is not an AuthorizeResponse");
+          }
+
+          if (response.getIdTagInfo().getStatus() == AuthorizationStatus.ACCEPTED) {
+            System.out.println("Authorization Accepted...");
+            System.out.println("Proceeding Transaction...");
```
pglob, line 51: Grammar:
- There should be a `with` after `Proceeding`

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+   * and then stop charging
+   *
+   * @param idTag id of user
+   */
+  public void StopCharging(String idTag) {
+    if (this.idTag == idTag) {
+      stopHandler.initiateStopTransaction(this.transactionId, idTag);
+      stateMachine.transition(stopHandler.getStateMachine().getCurrentState());
+    } else {
+      preAuthorize(this.transactionId, idTag);
```
pglob, line 101: `transactionId` is passed into `preAuthorize` as a `connectorId` parameter. While it doesn't have an effect, it is confusing. Either add an extra parameter for it or use `null`.

pglob (approved):
Just some minor things to address, should be good after that.

backend/src/main/java/com/sim_backend/transactions/TransactionHandler.java
JayQuack: I think I need a extra transition for a stateMachine in TransactionHandler. Otherwise when startHandler stateMachine turns into Charging stateMachine in TransactionHandler remains Preparing. I'll put TODO here to decide later.

pglob: What I mean is that `stateMachine` in `TransactionHandler` and `stateMachine` in the start/stop handlers refer to the same object in memory. Objects are passed by reference in java

PR #61, opened 2025-02-04
Update dummy server to specify message sequences

pglob:
This is a rework of the dummy server used in integration testing. It splits the `DummyServer.js` monolith into separate files for its WebSocket server and HTTP server. The dummy server can be ran, along with the frontend/backend, using `make dummy_server`.

Notable features:
- Normal mode: when the environment variable `NORMAL_MODE` exists, the dummy server will automatically respond to incoming OCPP messages. This makes some informal testing and demoing more simple. These predefined responses are in `dummy_server/NormalModeResponses.json`
- Response files: if a Json file is sent to `dummy-server/api/uploadResponse`, the WebSocket server will use it to determine what responses to send to a sequence of messages. Example:
>This input
>{
>   "a1":{"status":"Accepted", "currentTime":"X", "interval":5},
>    "b5":{"currentTime":"X"}
>}
>
>  Generate the following sequence of responses:
>  [3, "<message response id here>",{"status":"Accepted", "currentTime":"<timestamp added here>", "interval":5}]
>  5x  [3, "<message response id here>",{"currentTime":"<timestamp added here>"}]

The StatelessMessages test is not updated for this rework, and gets deprecated in the other integration testing pr. The reason for this is that the "Send Messages" button will very likely receive changes in the near future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,13 +7,13 @@ docker_debug:
 	DEBUG_STAGE=suspend-debug docker-compose --profile core up --build
 
 integration_test:
-	DEBUG_STAGE=run-debug docker-compose --env-file .env.internal --profile integration-test up --build
+	docker-compose --env-file .env.internal --profile integration-test up --build
 
 dummy_server:
-	docker compose --profile core --profile dummy-server up --build
+	NORMAL_MODE=enabled docker compose --profile core --profile dummy-server up --build
 
 dummy_server_debug:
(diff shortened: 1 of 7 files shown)


sille369:
Had tried "make dummy server" and it looks good to me>

sille369 (approved):
Everything looks good to me!

PR #62, opened 2025-02-04
Frontend updates for MVP

pglob:
* Connect charger labels to backend endpoints
* Add tests for labels
* Add Reboot button (tested via integration test in later PR)
* Create start/stop charging button (replaced "Plug in Vehicle") (tested via integration test in later PR)
* Formatted styles.css a bit + added styles
* Various other fixes

The screenshots below only really work out with the `MVP/simulator_and_transactions` branch as the backend does not yet support the Reboot button and live updating of the charger labels.
Screenshot 2025-02-03 at 8 26 09 PM

Screenshot 2025-02-03 at 8 26 16 PM

Screenshot 2025-02-03 at 8 26 20 PM

Screenshot 2025-02-03 at 8 26 45 PM

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -1,25 +1,47 @@
 import React, { useEffect, useState } from 'react';
 import ChargerFrame from './components/ChargerFrame';
 import ShowLogMessages from './components/ShowLogMessage';
+import './styles/styles.css';
 
 function App() {
-  const [message, setMessage] = useState('');
+  const [status, setStatus] = useState('loading');
 
-  // Hook to fetch data from backend when the component mounts
   useEffect(() => {
     fetch(`${process.env.REACT_APP_BACKEND_URL}/api/test`)
-      .then((response) => response.text())
-      .then((data) => setMessage(data))
-      .catch((error) => console.error('Error fetching test message:', error));
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error('Network response was not ok');
+        }
+        return response.text();
+      })
+      .then((data) => {
+        if (data === 'Ok') {
+          setStatus('success');
+        } else {
+          setStatus('failure');
+        }
+      })
+      .catch((error) => {
+        console.error('Error fetching test message:', error);
+        setStatus('failure');
+      });
   }, []);
 
-  // Render the component
   return (
     <div>
-      <h1>Frontend</h1>
-      <p data-testid="message">Message from backend: {message}</p>
-      <ChargerFrame />
-      <ShowLogMessages />
+      <h1>OCPP Charger Simulator</h1>
+      {status === 'failure' && (
+        <b data-testid="message">
+          ERROR: Backend at {process.env.REACT_APP_BACKEND_URL} was not
+          reachable
+        </b>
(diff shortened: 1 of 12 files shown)


frontend/src/App.js
zhihw, line 1: for the title"OCPP Charger Simulator"

frontend/src/App.js
```diff
@@ -2,21 +2,38 @@ import React, { useEffect, useState } from 'react';
...
-      <h1>Frontend</h1>
-      <p data-testid="message">Message from backend: {message}</p>
+      <h1>OCPP Charger Simulator</h1>
+      {status === 'failure' && (
+        <b data-testid="message">
+          ERROR: Backend at {process.env.REACT_APP_BACKEND_URL} was not
+          reachable
+        </b>
+      )}
       <ChargerFrame />
```
zhihw, line 37: for ChargerFrame components,maybe we should add a container so that we can adjust the styles after we add more chargers

pglob: @zhihw 
I added a `charger-frames-container` and another div. Copying that div 3 times gives the following:
Screenshot 2025-02-07 at 5 52 33 PM

zhihw (approved):
I checked all the UI design parts, and everything is good at the moment. I think I will continue to adjust the css and chargerframe in the next task. I am still studying pollChargerData. If the state value is updated every second, will it affect the user experience?

pglob:
> I checked all the UI design parts, and everything is good at the moment. I think I will continue to adjust the css and chargerframe in the next task. I am still studying pollChargerData. If the state value is updated every second, will it affect the user experience?

@zhihw, if a state is active for less than a second, it may not be visible. This only affects the ‘preparing’ state. We could probably poll it more often, I can look into what the overhead for that would be.

zhihw:
Sorry, it's my mistake. I initially thought that the state didn't have to be faster than other label items. I misunderstood.

zhihw:
> > I checked all the UI design parts, and everything is good at the moment. I think I will continue to adjust the css and chargerframe in the next task. I am still studying pollChargerData. If the state value is updated every second, will it affect the user experience?
> 
> 
> 
> @zhihw, if a state is active for less than a second, it may not be visible. This only affects the ‘preparing’ state (our other states last longer). We could probably poll it more often, I can look into what the overhead for that would be. 

Sorry, it's my mistake. I initially thought that the state didn't have to be faster than other label items. I misunderstood.

frontend/src/__test__/App.test.js
```diff
@@ -1,29 +1,47 @@
...
 import App from '../App';
 
-beforeEach(() => {
-  global.fetch = jest.fn(() =>
-    Promise.resolve({
-      text: () => Promise.resolve('Hello from the backend!'), // Mocked response
-    })
-  );
+// Ensure the backend URL is defined for testing
+beforeAll(() => {
```
zhihw, line 7: We should put hooks in describe to make the code more reasonable and consistent. Since we have modularized the tests, maybe we should not use hooks scoped to the whole file.

zhihw (approved):

PR #64, opened 2025-02-04
Rework Simulator for Thread Safety + Rebooting

pglob:
Rework simulator for modularity, rebooting, and thread safety
* Also adds transaction, electrical, and state related endpoints

Add thread safety to transactions
* Fix bug with setting transactionId
* Makes transactions thread safe
* Also add more comments

This really should be 2 PRs but they got very intertwined overtime.


backend/src/main/java/com/sim_backend/Main.java
```diff
@@ -3,58 +3,36 @@
...
-    // Register observers with the Websocket client
-    wsClient.onReceiveMessage(BootNotificationResponse.class, bootObserver);
-
-    // Boot the charger
-    stateMachine.transition(SimulatorState.BootingUp);
-
-    // Run the main simulator loop
-    SimulatorLoop.runSimulatorLoop(wsClient);
+    // Start the simulator
+    sim.Boot();
```
pglob, line 27: Functionality removed from `Main.java`, has been moved into `Simulator.Boot()` (found in `Simulator.java`)

pglob:
Semantics question for you guys: does `Simulator` or `Charger` make more sense? I initially named the new class `Simulator`, but that might be confusing since the whole application is a simulator.

Xerasin (approved):
I feel like `Charger` makes a lot more sense but it's up to you

PR #66, opened 2025-02-06
Field validity enforcement

cfarhan:
Jakarta validation constraints used to create non-intrusive field validity enforcement. Jakarta does NOT throw automatically upon constraint violations. Instead, the message validator class can be used to validate any OCPPMessage. This helps reduce the amount of refactoring needed to implement validation. It also allows for specific constraint violations (and violation output) to be managed within the OCPPMessage class itself.

Ensures all OCPPMessages adhere to their JSON schema constraints
- All _required_ JSON properties are checked to ensure they are non-null.
- All JSON string properties must adhere to their respective max string length.
- All validation errors can be outputted via MessageValidator.log_message(OCPPMessage).

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -97,6 +97,23 @@
             <artifactId>jakarta.validation-api</artifactId>
             <version>3.0.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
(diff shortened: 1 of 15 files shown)


pglob (approved):
Looks good, thanks!

PR #69, opened 2025-02-10
Fix rendering nested message payloads

pglob:
Messages with nesting in their payloads, like "idTagInfo" fields, would throw an error previously.

Also increase the max number of logged messages. I ran into a case where seeing an older message was useful, so I doubled it.


sille369 (approved):
I had tested it. Looks good to me!

PR #73, opened 2025-02-15
Cscap 159/get configuration

JayQuack:
Get Configuration Request, Response, Observer and test added


JayQuack:
Spotless will be applied

backend/src/main/java/com/sim_backend/websockets/observers/GetConfigurationObserver.java
```diff
@@ -0,0 +1,96 @@
...
+     * TODO: Retrieve GetConfigurationMaxKeys from the Charge Point
+     */
+    int GetConfigurationMaxKeys = 5;
+
+    if (request.getKey() == null) {
+      configurationInfo.add(
+          new GetConfigurationResponse.Configuration(
+              "MeterValueSampleInterval",
+              String.valueOf(registry.getMeterValueSampleInterval()),
+              true));
```
pglob, line 59: These booleans should be `false`. They are not readonly as far as I know. Let me know if I'm missing something

pglob (commented):
Just a couple things.

backend/src/main/java/com/sim_backend/websockets/observers/GetConfigurationObserver.java
pglob: @JayQuack I think the booleans in the `else` block also need to be `false`

JayQuack: OCPP mentions about read only state, when I set it as false, would Charger make it as a  readonly later ?

pglob: @JayQuack whether a configuration is read only is defined by `Accessibility` in section 9.1.17 and 9.1.19 of the OCPP 1.6 spec. These 2 configurations are always `RW` (read and write), never `R` (readonly).

Both of them can be modified by `ChangeConfiguration` messages. Therefore, readonly will be false.

JayQuack: I see read-only doesn't mean it's readonly, it should contain read-only value

pglob (approved):

PR #75, opened 2025-02-17
Add message timeout handling

pglob:
For some observers, a message timeout would previously lock up the simulator and leave it in a bad state. This PR adds the onTimeout() function to the OCPPMessageListener interface. By default, onTimeout() does nothing.

The following timeout handling was added to observers:
Authorize (before StartTransaction): Resets state to "Available"
StartTransaction: Resets state to "Available"
StopTransaction: Nothing, the listener was removed since the central system cannot prevent a transaction from stopping
BootNotification: The BootNotification is now retried with priority

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
@@ -57,7 +57,7 @@ public class OCPPWebSocketClient extends WebSocketClient {
   public static final String MESSAGE_PACKAGE = "com.sim_backend.websockets.messages";
 
   /** The OCPP Message Queue. */
-  private final MessageQueue queue = new MessageQueue();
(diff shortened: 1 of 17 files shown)


backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -336,6 +336,21 @@ public boolean pushMessage(final OCPPMessage message) {
     return queue.pushMessage(message);
   }
 
+  /**
+   * Add a OCPPMessage to the front of our send queue.
+   *
+   * @param message the message to be sent.
+   */
+  public boolean pushPriorityMessage(final OCPPMessage prioMessage) {
```
Xerasin, line 344: Probably move this method into the MessageQueue class, and pass it through like with what pushMessage does.

Xerasin (approved):

PR #78, opened 2025-02-18
Rework message dropdown

sille369:
* removed the boot notification button from the send message drop down menu
* reworked "send message button", made heartbeat, authorize buttons to be horizontal and covered only the send message button.


pglob (approved):
Revisit this as part of the "Make UI consistent" task. It would be good to have the lower 4 buttons all similar and neatly arranged.

PR #79, opened 2025-02-20
 CSCAP-148&CSCAP-149

zhihw:
I updated the route in MessageController to support multiple chargers in the form of /api/:chargerId/... In Main.java, I created 3 Charger instances and passed them to MessageController in array form, and modified the URL requested by the front end. Now the UI has three chargerframes. I have a problem now, and I haven't found the reason yet. The front end seems unable to communicate with the back end, and even api/test is not reachable.

### DIFF
--- a/backend/src/main/java/com/sim_backend/Main.java
+++ b/backend/src/main/java/com/sim_backend/Main.java
@@ -31,7 +31,6 @@ public final class Main {
       charger.Boot();
     }
   }
-  
 
   /**
    * Load configuration from environment variables or defaults.
--- a/backend/src/main/java/com/sim_backend/Main.java
+++ b/backend/src/main/java/com/sim_backend/Main.java
@@ -18,14 +18,20 @@ public final class Main {
     Javalin app = initializeFrontendAPI(config);
 
     // Create the charger
-    Charger charger = new Charger();
+    Charger charger1 = new Charger();
+    Charger charger2 = new Charger();
+    Charger charger3 = new Charger();
+    Charger [] chargers = new Charger[] {charger1, charger2, charger3};
 
     // Register REST API controllers and routes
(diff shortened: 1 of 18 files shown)


pglob:
@zhihw I think I might have sent you old documentation. It looks like the version of javalin we use has different syntax for variable paths. Swap out `/api/:chargerId/...` for `/api/{chargerId}/...`

zhihw:
> @zhihw I think I might have sent you old documentation. It looks like the version of javalin we use has different syntax for variable paths. Swap out `/api/:chargerId/...` for `/api/{chargerId}/...`

It worked! Thanks for your help!

pglob (approved):
Looks good! One thing, the 3 chargers all say `Charger A`, could you change that to `Charger A   Charger B     Charger C` or  `Charger 1   Charger 2     Charger 3`?

zhihw:
> Looks good! One thing, the 3 chargers all say `Charger A`, could you change that to `Charger A Charger B Charger C` or `Charger 1 Charger 2 Charger 3`?

Done, what do you think about the conflict?

pglob:
> > Looks good! One thing, the 3 chargers all say `Charger A`, could you change that to `Charger A Charger B Charger C` or `Charger 1 Charger 2 Charger 3`?
> 
> Done, what do you think about the conflict?

The changes made in MessageController.java are important for thread safety. Remove `if (!checkTransactionHandler...`, `if (!checkElec...`, and `if (!checkWsClient...` lines from your changes and keep what changes were made in main. `private boolean checkElec(Charger charger, Context ctx) {` on line 68 just got moved down a few lines to make room for the new `checkStateMachine` function.

The changes in Button.js were to remove the BootNotification button. If you remove `new BootButton(chargerID),` you can use your changes.

PR #81, opened 2025-02-20
Cscap 143/respond with call errors

Xerasin:
I plan to add a couple more once my other PR gets merged but otherwise I'm not too sure what else to add


pglob (changes requested):
Add a generic CallError near this line `log.error("Received Bad OCPP Message: ", exception);`. (Maybe `InternalError`?). That way, if there is an exception internally in a class that bubbles up, the central system will always be notified.

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -172,70 +175,130 @@ public void onMessage(String s) {
...
-          log.warn("Received OCPP error message with an unknown ID {}: {}", msgId, s);
-          throw new OCPPCannotProcessMessage(s, msgId);
+          this.queue.clearPreviousMessage(prevMessage);
+          OCPPMessageInfo info = prevMessage.getClass().getAnnotation(OCPPMessageInfo.class);
+          messageName = info.messageName() + "Response";
+          messageType = info.messageName();
+          this.recordRxMessage(s, messageType);
+          data = array.get(PAYLOAD_INDEX - 1).getAsJsonObject();
         }
+        case OCPPMessage.CALL_ID_ERROR -> {
```
pglob, line 235: For each of the 3 cases, add a comment and a blank line between them, or move the individual case handling into separate functions. It's getting hard to parse

pglob (approved):
Just address this, and the 1 above, and it's good to go

PR #82, opened 2025-02-23
Create initial implementation of charging profile

cfarhan:
If possible I'd like some feedback on this one @pglob , going to focus on finishing the other tickets so this doesn't bog me down. I'm honestly a little stuck on the implementation for this one, not sure which fields need to be included and what not. 

For now, I have the electrical transition just copying the charging profile if the few conditions are met.

### DIFF
--- a/backend/src/main/java/com/sim_backend/charger/Charger.java
+++ b/backend/src/main/java/com/sim_backend/charger/Charger.java
@@ -6,19 +6,12 @@ import com.sim_backend.state.ChargerState;
 import com.sim_backend.state.ChargerStateMachine;
 import com.sim_backend.transactions.TransactionHandler;
 import com.sim_backend.websockets.OCPPWebSocketClient;
-import com.sim_backend.websockets.enums.ChargePointErrorCode;
-import com.sim_backend.websockets.enums.Reason;
 import com.sim_backend.websockets.observers.BootNotificationObserver;
-import com.sim_backend.websockets.observers.ChangeAvailabilityObserver;
 import com.sim_backend.websockets.observers.ChangeConfigurationObserver;
 import com.sim_backend.websockets.observers.GetConfigurationObserver;
-import com.sim_backend.websockets.observers.MeterValuesObserver;
-import com.sim_backend.websockets.observers.StatusNotificationObserver;
-import com.sim_backend.websockets.observers.TriggerMessageObserver;
 import java.net.URI;
 import java.util.concurrent.locks.ReentrantLock;
 import lombok.Getter;
-import lombok.Setter;
 
 /**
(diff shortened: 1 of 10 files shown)


pglob (commented):
To get our `ElectricalTransition` to abide by a charging profile, we need the following:

At a high level, the `currentOffered` parameter gets set to `limit` (from `ChargingSchedulePeriod`). Everything else in a `ChargingProfile` and its subclasses just inform `ElectricalTransition` when to apply that limit.

Looking at the `ChargingSchedulePeriod` subclass:
- `startPeriod`- this one would need to be implemented as written. This time is relative to `startSchedule` from the `ChargingSchedule` class.
- `limit`- implemented as written.
- `numberPhases`- this is optional and our simulator is simulating 1 phase. Since the default is 3, we can ignore this field.

Looking at the `ChargingSchedule` subclass:
- `duration`- implemented as written.
- `startSchedule`- implemented as written.
- `chargingRateUnit`- as written, note that we ignore the number of phases here again and just do 1 phase.
- `chargingSchedulePeriod`- implemented as written.
- `minChargingRate`- Given this from the spec, "The optional charging schedule field minChargingRate may be used by the Charge Point to optimize the power distribution between the connectors. ", we can mostly ignore this parameter. Since we have 1 connector it does not hold much purpose. We just need to confirm that `minChargingRate` < `limit`s and throw an exception if not.

Looking at the `ChargingProfile` class:
- `chargingProfileId`- implemented as written. This one only actually comes into play when `5.16. Set Charging Profile` is implemented.
- `transactionId`- implemented as written.
- `stackLevel`-  implemented as written. I stated earlier that we may not have time for this one. However, looking over the spec again, it looks like less work than I anticipated. I think it should be achievable.
- `chargingProfilePurpose`-  implemented as written.
- `recurrencyKind`- implemented as written. This one is a bit weird with our simulator. I think something that could be done here is if the current simulator time is 24 hours/ 7 days after `startSchedule`, just add 24 hours/7 days to `startSchedule`.
- `validFrom`- implemented as written.
- `validTo`- implemented as written. Note that a profile that is past its valid time, is considered invalid, even if it is intended to have recurrence.
- `chargingSchedule`- implemented as written.


Given all the above, there needs to be a reference to the `TransactionHandler`. Note that `TransactionHandler.startInProgress` and `TransactionHandler.stopInProgress` are not indicators of if a transaction is currently in progress. If our state is `Charging`, we are guaranteed to have a transaction in progress.

Once a charging profile is assigned, we can just check the profiles/schedule whenever `currentOffered` is accessed. In effect, unless we call a function that accesses `currentOffered`, the charging profile does not exist. This means we do not need to check any time information independent of charging.

PR #83, opened 2025-02-23
Cscap 160/status notification

JayQuack:
It became a spaghetti code real quick.
I try to make it simple as possible.

it passed the MessageControllerTest.


backend/src/main/java/com/sim_backend/charger/Charger.java
```diff
@@ -44,6 +45,9 @@ public class Charger {
   /** A lock to ensure that only one Boot() or Reboot() operation can run at a time */
   private final ReentrantLock bootRebootLock = new ReentrantLock();
 
+  /** The Observer for StatusNotification */
+  private StatusNotificationObserver statusNotificationObserver;
```
pglob, line 49: We don't need to hold the reference to the observer. When we reboot, observers lose their references and are automatically garbage collected. This prevents any issues when rebooting in the middle of an onMessage for example.

The status notification observer can be created inside `Boot()`

backend/src/main/java/com/sim_backend/websockets/observers/StatusNotificationObserver.java
```diff
@@ -0,0 +1,82 @@
...
+import com.sim_backend.websockets.enums.ChargePointErrorCode;
+import com.sim_backend.websockets.enums.ChargePointStatus;
+import com.sim_backend.websockets.messages.StatusNotification;
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.AllArgsConstructor;
+
+/** Observer that handles StatusNotification requests when charger state changes. */
+@AllArgsConstructor
```
pglob, line 15: Register the observer for `onStateChanged` automatically in the constructor (like in the `BootNotificationObserver`).

backend/src/main/java/com/sim_backend/websockets/observers/StatusNotificationObserver.java
```diff
@@ -0,0 +1,82 @@
...
+public class StatusNotificationObserver implements StateObserver {
+
+  private final OCPPWebSocketClient client;
+  private List<StatusNotification> statusNotificationQueue = new ArrayList<>();
```
pglob, line 19: Since we only send the last statusNotification after coming online, this can reference a single `StatusNotification` object instead of a list.

backend/src/main/java/com/sim_backend/websockets/observers/StatusNotificationObserver.java
```diff
@@ -0,0 +1,82 @@
...
+  private ChargePointStatus ChargerStateToStatus(ChargerState state) {
+    return switch (state) {
+      case ChargerState.Available -> ChargePointStatus.Available;
+      case ChargerState.Preparing -> ChargePointStatus.Preparing;
+      case ChargerState.Charging -> ChargePointStatus.Charging;
+      default -> null;
+    };
+  }
+
+  /** Called in client goOline() function. */
```
pglob, line 74: Typo: `goOnline()`

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -79,6 +80,10 @@ public class OCPPWebSocketClient extends WebSocketClient {
   /** List to store received messages. */
   private final List<String> rxMessages = new CopyOnWriteArrayList<>();
 
+  /** StatusNotification Observer */
+  private final StatusNotificationObserver statusNotificationObserver =
```
pglob, line 84: Creating another `StatusNotificationObserver` will change the functionality. `statusNotificationObserver.onClientGoOnline();` won't call the same object that is in `Charger`.

pglob: Add a `StatusNotificationObserver` to the constructor. In `Charger.Boot()`, pass the observer in when creating the websocket client. This way we have a single `StatusNotificationObserver`.

backend/src/main/java/com/sim_backend/websockets/observers/StatusNotificationObserver.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  @Override
+  public void onStateChanged(ChargerState newState) {
+    ChargePointStatus status = ChargerStateToStatus(newState);
+
+    if (status == null) {
+      return;
+    }
+    ZonedDateTime timestamp = client.getScheduler().getTime().getSynchronizedTime();
+    sendStatusNotification(1, ChargePointErrorCode.NoError, "", status, timestamp, "", "");
```
pglob, line 35: Confirm that when a field is "", it is not present in the message.

For example (X represents some data):
the payload SHOULD NOT be- `{"connectorId":X, "errorCode":"X", "info":"", "status":"X", "timestamp":"X", "vendorId":"" "vendorErrorCode":""}`
the payload SHOULD  be-  `{"connectorId":X, "errorCode":"X", "status":"X", "timestamp":"X"}`

pglob (changes requested):
Add some testing confirming that a StatusNotification is pushed when state changes occur + some offline testing.

JayQuack:
Can you please check if it's going to work?
I will test and message output when code is acceptable.

backend/src/main/java/com/sim_backend/charger/Charger.java
```diff
@@ -88,6 +90,11 @@ public void Boot() {
       GetConfigurationObserver getConfigurationObserver =
           new GetConfigurationObserver(wsClient, config);
 
+      statusNotificationObserver.setClient(wsClient);
```
pglob, line 93: I think it is possible for a thread issue to occur here. `statusNotificationQueue` may not be `null` when a reboot occurs. 

To prevent any thread issues from cropping up, create a `new StatusNotificationObserver` instance here. The old one will get garbage collected.

pglob (commented):
Alright, I think it looks functionally good. Just this one thing below

pglob:
If you add `"StatusNotification":{}` to `dummy_server/NormalModeResponses.json`, if you run the simulator with `dummy make`, the status notifications from changing states should appear in the log UI.

backend/src/main/java/com/sim_backend/charger/Charger.java
```diff
@@ -84,6 +90,9 @@ public void Boot() {
       GetConfigurationObserver getConfigurationObserver =
           new GetConfigurationObserver(wsClient, config);
```
pglob, line 92: Sorry for the back and forth. That line you had before (`statusNotificationObserver.setClient(wsClient);`) should be added back here.

backend/src/test/java/com/sim_backend/websockets/observers/StatusNotificationObserverTest.java
pglob, line 1: Add a test case that verifies a status notification message gets sent on state changes.

pglob (approved):

PR #85, opened 2025-02-24
Make UI consistent:

sille369:
* removed un-used states from button.js
* used font Arial global
* algined those buttons with styles

### DIFF
--- a/frontend/src/components/Button.js
+++ b/frontend/src/components/Button.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types'; // Import PropTypes
 import AuthorizeButton from '../components/buttons/AuthorizeButton';
 import HeartbeatButton from '../components/buttons/HeartbeatButton';
@@ -7,8 +7,6 @@ import '../styles/styles.css';
 
 // Implementation of the button component, which includes a drop-down menu and an online/offline button
 function Button({ chargerID, isOnline, isActive }) {
-  const [openDropdown, setOpenDropdown] = useState(false); // Track dropdown open/close state
-
   // Create instances of each button class
   const buttons = [
     new AuthorizeButton(chargerID),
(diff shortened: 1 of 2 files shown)


frontend/src/styles/styles.css
```diff
@@ -1,3 +1,7 @@
+* {
+  font-family: 'Inter', Arial, sans-serif;
```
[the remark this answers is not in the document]
pglob: With the font change, there's a lot of bold going on. To me, it is harder to read and strains the eyes a bit. Below is a way to reduce the boldness.
```
* {
  font-family: 'Inter', Arial, sans-serif;
  font-optical-sizing: auto;
  font-weight: 300; /* Arial fallback rounds up to 400 */
}

strong {
  font-weight: 601; /* Arial fallback rounds up to 700 */
}

h1 {
  font-weight: 700;
}
```

If we do the above, we would probably want to change the font-weight in `.round-rect-container` from `bold` to some lower value.
```
/* Styling for the round rectangle container for the message boxes header */
.round-rect-container {
  margin-top: 25px;
  /* margin-bottom: 5px; */
  background-color: #b3f4dc;
  padding: 10px 20px;
  border-radius: 20px;
  font-weight: bold;
  display: inline-block;
  color: #333;
}
```

We can also reduce border widths of some elements to 1px.

Let me know what you think. We can also explore other options if you'd like.

frontend/src/components/Button.js
pglob, line 1: There are a lot of different shades of colors etc. Define some properties for set colors like dark red, light red, light grey, white, etc.

pglob, line 1: Border's should be consistent as well. The log message border is grey but the charger frame border is black.

pglob, line 1: Temporarily add the `<ErrorMenu />` component into `App.js` just to check if the UI there is consistent too.

pglob (changes requested):

frontend/src/styles/styles.css
```diff
@@ -92,22 +100,41 @@
   cursor: pointer;
 }
 
-/* Darken a button when pressed */
-button:active {
```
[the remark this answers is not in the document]
sille369: Added back.

pglob (approved):
Looks good.

1 last thing, the scrollbar on the log messages clips the border. If we can round those scrollbar edges to match that would be great.

You're good to merge after fixing that.

PR #86, opened 2025-02-27
Fix logging errors

sille369:
* Fixed record Rx messages for CALL_ID_REQUEST
* Fixed GetConfiguration message crashed the frontend
* Fixed CallErrors messages parsing error.

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
+++ b/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
@@ -248,13 +248,10 @@ public class OCPPWebSocketClient extends WebSocketClient {
       this.receivedIDs.add(msgId);
 
       ParseResults results;
-
       int callId = array.get(CALL_ID_INDEX).getAsInt();
       switch (callId) {
         case OCPPMessage.CALL_ID_REQUEST -> results = this.parseOCPPRequest(json, msgId, array);
-
         case OCPPMessage.CALL_ID_RESPONSE -> results = this.parseOCPPResponse(json, msgId, array);
-
         case OCPPMessage.CALL_ID_ERROR -> {
           this.handleOCPPMessageError(json, msgId, array);
           return;
@@ -328,6 +325,7 @@ public class OCPPWebSocketClient extends WebSocketClient {
     }
 
     String messageName = array.get(NAME_INDEX).getAsString();
+    this.recordRxMessage(json, messageName);
     return new ParseResults(messageName, array.get(PAYLOAD_INDEX).getAsJsonObject());
   }
 
@@ -365,7 +363,6 @@ public class OCPPWebSocketClient extends WebSocketClient {
 
     this.queue.clearPreviousMessage(prevMessage);
     OCPPMessageInfo info = prevMessage.getClass().getAnnotation(OCPPMessageInfo.class);
-
     String messageName = info.messageName() + "Response";
     this.recordRxMessage(json, info.messageName());
     return new ParseResults(messageName, array.get(PAYLOAD_INDEX - 1).getAsJsonObject());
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -340,7 +338,6 @@ private ParseResults parseOCPPResponse(String json, String msgId, JsonArray arra
 
     this.queue.clearPreviousMessage(prevMessage);
     OCPPMessageInfo info = prevMessage.getClass().getAnnotation(OCPPMessageInfo.class);
-
```
pglob, line 343: I say leave this space in, it helps readability

frontend/src/components/ShowLogMessage.js
```diff
@@ -93,7 +93,33 @@ const ShowLogMessages = ({ chargerID }) => {
   const parseMessage = (message) => {
     try {
       const parsedMessage = JSON.parse(message); // Parse the string into an array
-      const [TimeStamp, NumId, messageId, messageName, payload] = parsedMessage;
+      let TimeStamp, NumId, messageId, messageName, payload, description;
```
pglob, line 96: Instead of specifically fixing messages with a `configurationKey` field, investigate what part of the json structure caused the previous failure. Then make sure we are able to parse messages with that structure.

The current implementation seems like a bandaid fix to me.

frontend/src/components/ShowLogMessage.js
```diff
@@ -144,14 +188,26 @@ const ShowLogMessages = ({ chargerID }) => {
...
+
+      //CALL_ID_REQUEST
+      if (parsedMessage.length === 6) {
+        [messageName, TimeStamp, NumId, messageId, type, payload] =
+          parsedMessage;
+        console.log('Request received message: {}', type);
+      }
+
+      //CALL_ID_RESPONSE
+      if (parsedMessage.length === 5) {
```
pglob, line 201: I don't think this length based implementation will work. It is possible for the central system to send us a call error as well (with a length of 5)

pglob (changes requested):

frontend/src/components/ShowLogMessage.js
sille369: on my recordRxMessage, I appended a timestamp and message type, if it's a callerror, it would be the length of 7

pglob: Ah I see. Add a brief comment explaining what happens for `CALL_ID_ERROR` messages

frontend/src/components/ShowLogMessage.js
sille369: the json type of Request and response are different (please take a look of the first and last sent messages), and the json type of the error one I asked previously, it didn’t get triggered by pressing the heartbeat button, i think that was the general error output. The backend handles request, response and error by the array length, and so far I don’t know how to get to know the json type of the error one. Like what you said, if we are able to parse messages with the same structure that would be beneficial for making the UI consistent.
Screen Shot 2025-02-25 at 11 38 43 PM

pglob: I think I need some clarification here. When you say json type, are you referring to the syntax/structure? Or are you referring to a specific field?

In this comment I am just referring to this error/failure that comes up from the trying to render the GetConfiguration response we send out.
Screenshot 2025-02-27 at 6 15 26 PM

The CallErrors we send out (resulting from the heartbeats) are structured the same way as CallError's that the central system would send us. So in theory, sending this with wscat, `[4,"<id>","ProtocolError","Some error message",{}]`, should be a CallError that goes in "Received Messages".

sille369: The json type I mean the data structures, like the screenshot above, the first (bootnotification) and the last(configurationKey) sent messages are in different structures.
And for the GetConfiguration case, I got the below response:
Screen Shot 2025-03-03 at 12 06 46 PM
Screen Shot 2025-03-03 at 12 09 38 PM
And for [4,"<id>","ProtocolError","Some error message",{}] in received message, I will implement it.

pglob (approved):

PR #89, opened 2025-02-28
CSCAP 189, Responses have random IDs

Xerasin:
Added a requirement to need the request to create a response object, also prevented the ability to received the same id twice.

### DIFF
--- a/backend/src/main/java/com/sim_backend/websockets/types/OCPPMessage.java
+++ b/backend/src/main/java/com/sim_backend/websockets/types/OCPPMessage.java
@@ -43,7 +43,7 @@ public abstract class OCPPMessage implements Cloneable {
    * @param messageID The wanted message ID.
    */
   protected OCPPMessage(String messageID) {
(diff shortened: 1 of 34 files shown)


backend/src/main/java/com/sim_backend/websockets/types/OCPPMessage.java
```diff
@@ -37,6 +37,15 @@ protected OCPPMessage() {
     this.messageID = generateMessageID();
   }
 
+  /**
+   * The constructor for an OCPP message given a wanted message id.
+   *
+   * @param messageID The wanted message ID.
+   */
+  protected OCPPMessage(String messageID) {
+    this.messageID = generateMessageID();
```
pglob, line 46: Should this be `this.messageID = messageID`?

Xerasin: woops lol

pglob (approved):

PR #92, opened 2025-03-02
Cscap 167/trigger message

zhihw:


backend/src/main/java/com/sim_backend/websockets/messages/TriggerMessageResponse.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Getter
+@EqualsAndHashCode(callSuper = true)
+@OCPPMessageInfo(
+    messageCallID = OCPPMessage.CALL_ID_RESPONSE,
+    messageName = "TriggerMessageResponse")
+public class TriggerMessageResponse extends OCPPMessageResponse implements Cloneable {
+  private TriggerMessageStatus status;
```
pglob, line 17: Add a not null check here

backend/src/main/java/com/sim_backend/websockets/observers/TriggerMessageObserver.java
```diff
@@ -0,0 +1,123 @@
...
+    switch (state) {
+      case Available:
+        return ChargePointStatus.Available;
+      case BootingUp:
+        return ChargePointStatus.Preparing;
+      case Preparing:
+        return ChargePointStatus.Preparing;
+      case Charging:
+        return ChargePointStatus.Charging;
+      case PoweredOff:
```
pglob, line 117: If we are PoweredOff or BootingUp, send a reject response. Technically it shouldn't be possible to hit those cases (it goes against the OCPP spec), so this is a just in case thing.

Another reason for this is that `Preparing` refers to the state right before starting a charge, and `Unavailable` refers to a state where we have booted up and connected, but are not allowing charges to start.

backend/src/main/java/com/sim_backend/websockets/observers/TriggerMessageObserver.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private StatusNotification createStatusNotification(Integer connectorId) {
+    if (connectorId == null) {
+      connectorId = 0;
+    }
+    ChargePointErrorCode errorCode = ChargePointErrorCode.NoError;
+    String info = "";
+    ChargePointStatus status = mapStateToChargePointStatus(stateMachine.getCurrentState());
+    ZonedDateTime timestamp = ZonedDateTime.now();
```
pglob, line 100: Use `webSocketClient.getScheduler().getTime().getSynchronizedTime()` here. This function gives the internal simulator time that reflects the time of the central system.

backend/src/main/java/com/sim_backend/websockets/observers/TriggerMessageObserver.java
```diff
@@ -0,0 +1,123 @@
...
+        break;
+      case MeterValues:
+        // TODO implement MeterValues trigger action here
+        responseStatus = TriggerMessageStatus.NotImplemented;
+        break;
+      case StatusNotification:
+        triggeredAction =
+            () -> {
+              StatusNotification statusNotification =
+                  createStatusNotification(triggerMessage.getConnectorId());
```
pglob, line 76: If the connector id is not 1 or 0, reject the trigger. Since we only allow for a single connector in our simulator, it's best to not give the impression we support multiple.

backend/src/main/java/com/sim_backend/websockets/observers/TriggerMessageObserver.java
```diff
@@ -0,0 +1,123 @@
...
+        } else {
+          triggeredAction =
+              () -> {
+                new BootNotificationObserver(webSocketClient, stateMachine)
+                    .handleBootNotificationRequest();
+              };
+        }
+        break;
+      case DiagnosticsStatusNotification:
+        responseStatus = TriggerMessageStatus.Rejected;
```
pglob, line 60: Not implemented would be better for this and FirmwareStatusNotification.

backend/src/main/java/com/sim_backend/websockets/observers/TriggerMessageObserver.java
```diff
@@ -0,0 +1,123 @@
...
+    webSocketClient.onReceiveMessage(TriggerMessage.class, this);
+  }
+
+  @Override
+  public void onMessageReceived(OnOCPPMessage message) {
+    if (!(message.getMessage() instanceof TriggerMessage triggerMessage)) {
+      throw new ClassCastException("Received message is not a TriggerMessage");
+    }
+
+    if (!MessageValidator.isValid(triggerMessage)) {
```
pglob, line 39: This validation was recently moved into central locations and can be safely removed from the observer.

PR #93, opened 2025-03-03
CSCAP-175 - ChangeAvailabilityObserver

Xerasin:
Added a version of the observer though I'm not sure this implementation is correct, but I filled it out to the best of my understanding.


backend/src/main/java/com/sim_backend/websockets/messages/ChangeAvailability.java
```diff
@@ -0,0 +1,22 @@
...
+import com.sim_backend.websockets.enums.AvailabilityType;
+import com.sim_backend.websockets.types.OCPPMessageRequest;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class ChangeAvailability extends OCPPMessageRequest implements Cloneable {
+  @SerializedName("connectorId")
+  private int connectorID;
```
pglob, line 13: Add some validation, like connectorID >= 0, type not null etc

backend/src/main/java/com/sim_backend/state/ChargerStateMachine.java
```diff
@@ -13,13 +13,20 @@ public class ChargerStateMachine {
...
-              Set.of(ChargerState.Charging, ChargerState.Available, ChargerState.Faulted),
-          ChargerState.Charging, Set.of(ChargerState.Available, ChargerState.Faulted),
-          ChargerState.Faulted, Set.of(ChargerState.Available));
+          Set.of(ChargerState.Charging, ChargerState.Available, ChargerState.Faulted),
+          ChargerState.Charging,
+          Set.of(ChargerState.Available, ChargerState.Faulted),
+          ChargerState.Faulted,
+          Set.of(ChargerState.Available),
+          ChargerState.Unavailable,
+          Set.of(ChargerState.Available));
```
pglob, line 29: Unavailable should be able to transition into a `Faulted` state

backend/src/main/java/com/sim_backend/websockets/observers/ChangeAvailabilityObserver.java
```diff
@@ -0,0 +1,82 @@
...
+    stateMachine.addObserver(this);
+  }
+
+  private void changeAvailability(ChargerState newState) {
+    ChargePointStatus chargePointStatus =
+        newState == ChargerState.Unavailable
+            ? ChargePointStatus.Unavailable
+            : ChargePointStatus.Available;
+
+    StatusNotification statusNotification =
```
pglob, line 36: The status notification observer should handle this message sending. Just update its `ChargerStateToStatus` function to include `Unavailable`

backend/src/main/java/com/sim_backend/websockets/observers/ChangeAvailabilityObserver.java
```diff
@@ -0,0 +1,82 @@
...
+        new StatusNotification(
+            1,
+            ChargePointErrorCode.NoError,
+            "Availability changed",
+            chargePointStatus,
+            null,
+            "",
+            "");
+    client.pushMessage(statusNotification);
+    stateMachine.transition(newState);
```
pglob, line 46: Use `checkAndTransition` here for safety, since we are expecting to be in `Available` or `Unavailable`.

pglob: The return value could also be used to reject the ChangeAvailability message if our simulator is in the wrong state

backend/src/main/java/com/sim_backend/websockets/observers/ChangeAvailabilityObserver.java
```diff
@@ -0,0 +1,82 @@
...
+            : ChargerState.Available;
+
+    AvailabilityStatus responseStatus = AvailabilityStatus.REJECTED;
+
+    if (stateMachine.getCurrentState() == newState) {
+      responseStatus = AvailabilityStatus.ACCEPTED;
+    } else if (stateMachine.inTransaction()) {
+      responseStatus = AvailabilityStatus.SCHEDULED;
+      wantedState = newState;
+    } else {
```
pglob, line 65: We should also make sure we don't change availability during non-OCPP states like `BootingUp` or `PoweredOff`. Technically a ChangeAvailability message should never be received in that state, but just in case.

backend/src/main/java/com/sim_backend/websockets/observers/ChangeAvailabilityObserver.java
pglob, line 1: Add javadocs

backend/src/main/java/com/sim_backend/charger/Charger.java
```diff
@@ -91,6 +92,8 @@ public void boot() {
           new ChangeConfigurationObserver(wsClient, config);
       GetConfigurationObserver getConfigurationObserver =
           new GetConfigurationObserver(wsClient, config);
+      ChangeAvailabilityObserver changeAvailabilityObserver =
```
pglob, line 95: Observers are garbage collected during a reboot (this is to prevent issues with old references and threads). For persistence with the wantedState, `Charger` should store some availability flag.

pglob (changes requested):

pglob (approved):
Looks good, just slap on some testing

PR #98, opened 2025-03-05
CSCAP-213 Fix Heartbeat issues

Xerasin:
I also added a method for registering a listener for when we push a message. I'll add testing soon.

 I had to change `host.docker.internal` to `dummy-server` to get this to run on my server box for testing, I don't believe `host.docker.internal` is a default docker setting.

### DIFF
--- a/backend/src/main/java/com/sim_backend/charger/Charger.java
+++ b/backend/src/main/java/com/sim_backend/charger/Charger.java
@@ -69,7 +69,7 @@ public class Charger {
   /** Constructs a new Charger instance */
   public Charger() {
     // TODO: Get central system URI from frontend or command line
-    this.config = new ConfigurationRegistry("temptag", "ws://dummy-server:9000");
+    this.config = new ConfigurationRegistry("temptag", "ws://host.docker.internal:9000");
   }
 
   /**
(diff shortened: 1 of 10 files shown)


pglob:
Is your server box on linux? `host.docker.internal` is default to Windows/Mac afaik. Linux has an extra requirement to get it working:
https://stackoverflow.com/questions/48546124/what-is-the-linux-equivalent-of-host-docker-internal/67158212#67158212

Xerasin:
> Is your server box on linux? `host.docker.internal` is default to Windows/Mac afaik. Linux has an extra requirement to get it working: https://stackoverflow.com/questions/48546124/what-is-the-linux-equivalent-of-host-docker-internal/67158212#67158212

It is! I added the fix on there, let me know if this breaks it on windows. idk if we should keep it on `host.docker.internal` if that's a WSL feature though, feels more appropriate to use the name of the docker container.

pglob (approved):
Looks good. The original reason for using `host.docker.internal` was that it works for both regular use, and the headless web browser that integration tests use.

pglob:
@Xerasin is this ready to be merged? I'd like it in before we record demos

PR #99, opened 2025-03-06
Added test cases for Rx/Tx messages

sille369:


backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
pglob, line 1: Looking more closely at `rxRequestNames`, is there a reason it needs to be a list? We are only every accessing the last element from it, so it could just be a single `String`.

pglob (changes requested):

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
sille369: I think we need it as a list, because we need to keep tracking of the the last Rx CallRequest. But if "Getconfiguration" is the only Rx CallRequest, then we don't need rxRequestNames as a list. otherwise, we need it as a list to track the last Rx CallRequest name.

pglob: I mean that the only time `rxRequestNames` is accessed is with `messageName = rxRequestNames.get(rxRequestNames.size() - 1);`. 

So if our `rxRequestNames` list is:
```
{GetConfiguration, ChangeAvailability, TriggerMessage}
```
those first 2 elements are not needed and will never be accessed. So we could store it with a single `String`

sille369: oh, I see.

backend/src/test/java/com/sim_backend/websockets/OCPPWebSocketClientTest.java
```diff
@@ -851,50 +851,43 @@ public void onMessageReceived(OnOCPPMessage message) {}
   }
 
   @Test
-  public void testrecordTxMessage() throws Exception {
-    client.rxRequestNames.add("GetConfiguration");
+  public void testrecordTxMessage_case1() throws Exception {
```
pglob, line 854: Have a more descriptive name than `case1` or `case2`

backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java
```diff
@@ -135,16 +135,31 @@ public static JsonArray insertElementAt(JsonArray jsonArray, int index, JsonElem
...
    */
   public void recordTxMessage(String message) {
+    if (message == null) {
+      log.error("Sent null message");
+      return;
+    }
+
     Gson gson = GsonUtilities.getGson();
     JsonArray array = gson.fromJson(message, JsonArray.class);
+    try {
```
pglob, line 145: If `JsonSyntaxException` is from `JsonArray array = gson.fromJson(message, JsonArray.class);`, the try catch block below won't do anything.

pglob: Same thing down at line 194

pglob (approved):
Just a couple things, then it's good to go.

PR #101, opened 2025-03-08
Adapted different screen sizes

sille369:


pglob (changes requested):
Ah sorry, when I listed those resolutions, I intended those to just be some common resolutions. We should pay extra attention to that range of resolutions. If someone if viewing this website with a screen that is much bigger or smaller, it’s less of a concern how it looks.

With this current implementation, any screen size that doesn’t exactly match will fallback to default. (1920px will look normal but 1925px will be too small)

frontend/src/styles/styles.css
pglob, line 1: I think it would be good to experiment with relative units instead of just px. Also maybe assuming the height is less than the full screen would also help (when determining what width/size values to use).

pglob: Here's an example of what I'm talking about. My screen is 1440x900, but if I'm not full screen, the sizing is a bit off.
Screenshot 2025-03-10 at 10 51 07 PM
In this case the Received messages are not visible at the same time as everything else.

pglob: In full screen, it's a little better:
Screenshot 2025-03-10 at 10 51 49 PM

pglob: This is what it looks like zoomed out, which I think is good (ignoring zoom artifacts):
Screenshot 2025-03-10 at 10 51 20 PM

frontend/src/styles/styles.css
```diff
@@ -86,6 +86,262 @@ h1 {
   position: relative; /* This allows absolute positioning of child elements */
 }
 
+/* Adapt Different Screen Sizes*/
```
pglob, line 89: Consider moving this to the top of the file

frontend/src/styles/styles.css
sille369: I thought since there is a scroll bar for the height, it should be okay. I will resize based on the height as well

sille369: I use the responsive screen size with web inspect, with the resolution 1440 X 900, it looks a little bit like your zoom out effect.
Screen Shot 2025-03-11 at 1 56 59 PM

pglob:
Summarizing what was talked about in the meeting:
- Assume that some of the screen height gets taken up by the browser tabs and a toolbar
    - Test with ~90% of the height of those specific resolution examples
- If the charger frame, all the sent messages, and at least 1 received message are visible at the same time, it’s good to go.
    - This probably requires adjusting the breakpoints, so the UI is slightly smaller.

pglob (changes requested):
Ok, I think we're getting there.

Also I forgot my screen size is actually 2560 × 1600, which may have caused differences between our views.

One more thing, please define an @media for sizes widths less than 1360px. Right now sizes under that go to whatever the lower CSS has defined in it

pglob:
Hang on on making changes, I'm looking into something.

pglob:
Alright, ran into an issue where at some screen sizes, the placement of text is weird. Like below, the text is in the bottom part of all the buttons instead of being centered. I imagine there is some kind of conflict with the text size and padding/margins.
Screenshot 2025-03-13 at 10 09 48 PM

pglob (approved):

PR #102, opened 2025-03-09
Remote Start and Stop Transactions

JayQuack:
RemoteStartTransaction Request, Response, Observer and test Added
RemoteStopTransaction Request, Response, Observer and test Added
I did it it one branch due to merge conflict issue I had last time.


backend/src/main/java/com/sim_backend/websockets/observers/RemoteStartTransactionObserver.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Processes incoming RemoteStartTransaction messages and handles Start Transaction.
+   *
+   * @param message the received OCPP message, expected to be a RemoteStartTransaction.
+   * @throws ClassCastException if the message is not a RemoteStartTransaction.
+   */
+  @Override
+  public void onMessageReceived(OnOCPPMessage message) {
```
pglob, line 30: Some extra handling is required. The ability to start a transaction is dependent on state (a bad starting state should result in rejecting the remote start). The `startCharing()` function will also authorize if the provided id tag is different. There is also the atomic flag that needs to be set when starting a charge.

It would probably be cleaner to handle this in an overloaded version of `startCharging()` that gets an auth flag.

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStopTransactionObserver.java
```diff
@@ -0,0 +1,48 @@
...
+  public void onMessageReceived(OnOCPPMessage message) {
+    if (!(message.getMessage() instanceof RemoteStopTransaction request)) {
+      throw new ClassCastException("Message is not an RemoteStopTransaction Request");
+    }
+
+    String status = "Rejected";
+    TransactionHandler transactionHandler = charger.getTransactionHandler();
+
+    if (transactionHandler.getTransactionId().get() == request.getTransactionId()) {
+      transactionHandler.stopCharging(null, Reason.REMOTE);
```
pglob, line 41: `stopCharging` can be reworked to return a bool. That bool can determine if we reject the request.

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStopTransactionObserver.java
```diff
@@ -0,0 +1,48 @@
...
+
+/** Observer that handles Getting RemoteStopTransaction requests and response. */
+public class RemoteStopTransactionObserver implements OnOCPPMessageListener {
+
+  private final OCPPWebSocketClient client;
+  private final Charger charger;
+
+  public RemoteStopTransactionObserver(OCPPWebSocketClient client, Charger charger) {
+    this.client = client;
+    this.charger = charger;
```
pglob, line 20: Same issue with having a reference to `charger` here.

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStopTransactionObserver.java
pglob: `stopCharging` may also result in authorization, which should not occur in this case. It probably needs an no auth version as well.

pglob (changes requested):

JayQuack:
I don't have this run fail problem for ChargingProfile, why would that be an issue for this commit ??

pglob:
@JayQuack the issue might be some cached build, try rebasing your branch onto main. There was some moving around of ChargingProfile classes, so the test runner may be not clearing its workspace as intended.

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStartTransactionObserver.java
```diff
@@ -31,16 +42,22 @@ public void onMessageReceived(OnOCPPMessage message) {
...
-      charger.getTransactionHandler().startCharging(request.getConnectorId(), request.getIdTag());
+
+      if (configurationRegistry.isAuthorizeRemoteTxRequests()) {
+
+        System.out.println(
+            "RemoteStartTransaction Authorization Required, Sending Authorization Request...");
+        transactionHandler.startCharging(request.getConnectorId(), request.getIdTag());
+      } else {
+        System.out.println("RemoteStartTransaction Request Received... Starting Transaction...");
+        stateMachine.checkAndTransition(ChargerState.Available, ChargerState.Preparing);
```
pglob, line 58: This doesn't send out a `StartTransaction` message like the spec requires. We already have the logic for this in `TransactionHandler`:
```
                startHandler.initiateStartTransaction(
                    connectorId, idTag, transactionId, elec, startInProgress);
                TransactionHandler.this.idTag = idTag;
```

Just create a public function in `TransactionHandler` that does the above. That way, we don't have to think about transaction management inside this observer (this keeps the class nice and isolated).

pglob: We also need this in that public function to properly check the state and be thread safe. Booleans should be returned here instead of void.

```
    if (stateMachine.getCurrentState() != ChargerState.Available || startInProgress.get()) {
      return;
    }

    // Attempt to set the startInProgress flag.
    if (!startInProgress.compareAndSet(false, true)) {
      return;
    }
```

pglob: And the state change should be also be checked for a return. Any bad state needs to result in us sending a rejected response for the remote start.

pglob (changes requested):

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStartTransactionObserver.java
pglob, line 1: Also if the message has a Charging Profile in it, we can add that with `electricalTransition .getChargingProfileHandler() .addChargingProfile(<profile here>);`

pglob: This functionality is defined in `5.16.2. Setting a charge profile in a RemoteStartTransaction request`. 

> If the Central System includes a ChargingProfile, the ChargingProfilePurpose MUST be set to TxProfile and the
transactionId SHALL NOT be set.

If the central systems violate the restriction above, send a call error detailing it.

pglob:
@JayQuack try rebasing instead of using merge, I can't tell what changes are yours and what were already on main.

PR #105, opened 2025-03-13
HTTP_API.md is placed in the project root directory

zhihw:


HTTP_API.md
```diff
@@ -0,0 +1,95 @@
...
+- **Functionality**: Sends a heartbeat message. Constructs a `Heartbeat` message to indicate that the charging station is still online, and pushes it to the backend via the WebSocket client.
+- **Response**: Returns `"OK"` on success.
+
+### GET `/api/{chargerId}/state`
+
+- **Functionality**: Retrieves the current state of the charging station. The endpoint extracts the current state from the state machine.
+- **Response**: Returns the current state as a string.
+
+### POST `/api/{chargerId}/state/status`
+
```
pglob, line 28: I'd add an example POST payload here.

HTTP_API.md
```diff
@@ -0,0 +1,95 @@
...
+- **Response**: Returns the value as a string.
+
+### GET `/api/{chargerId}/electrical/current-import`
+
+- **Functionality**: Retrieves the current import value. Returns the current value from the electrical component.
+- **Response**: Returns the value as a string.
+
+### GET `/api/{chargerId}/get-idtag-csurl`
+
+- **Functionality**: Retrieves the configuration parameters of the charging station.
```
pglob, line 89: These 2 endpoints only configure the idTag and central system URL, not any configuration parameter. We don't want a user thinking this is for OCPP configurations

HTTP_API.md
```diff
@@ -0,0 +1,95 @@
...
+- **Response**: Returns a JSON-formatted string.
+
+### POST `/api/{chargerId}/update-idtag-csurl`
+
+- **Functionality**: Updates the configuration parameters of the charging station.
```
pglob, line 94: These 2 endpoints only configure the idTag and central system URL, not any configuration parameter. We don't want a user thinking this is for OCPP configurations

I'd also add an example POST payload here.

pglob (approved):

PR #107, opened 2025-03-13
CSCAP-212 - CallError Issues

Xerasin:
I removed the callerrors from response but I don't know if I should remove all the "generic" ones as well. Also, I wasn't able to replicate the but you wrote but I fixed an issue that may of been causing it.


pglob:
@Xerasin the way I read the spec, there are supposed to be zero CallErrors in response to a CallResponse. We should probably remove all the `pushCallError` calls from response handling as well for cleanliness. These locations should just have log output instead or something

pglob (approved):

PR #110, opened 2025-03-13
Config

zhihw:
It can read command line parameters and keep the configuration when the container is restarted via update config properties file when configuration changes. At present, I have added environment variables for command line parameters in the backend Dockerfile. You can simulate command line parameter changes by modifying the environment variables.


pglob:
@zhihw Is there a way right now of passing in the variables when running `make docker` or `make dummy_server`?

zhihw:
im not sure but It seems that it is not possible. Make docker is only
building the image.

backend/src/main/java/com/sim_backend/config/ConfigurationRegistry.java
pglob, line 1: Please add some basic unit testing for this file, then it’s good to go.

pglob (approved):

PR #111, opened 2025-03-14
Create SetChargingProfileObserver and test

cfarhan:
Creates the SetChargingProfileObserver class.

### DIFF
--- a/backend/src/main/java/com/sim_backend/electrical/ChargingProfileHandler.java
+++ b/backend/src/main/java/com/sim_backend/electrical/ChargingProfileHandler.java
@@ -49,7 +49,7 @@ public class ChargingProfileHandler {
    * Adds a new charging profile to the appropriate list based on its purpose. Ensures profiles with
    * the same stack level and purpose do not duplicate.
    */
-  public void addChargingProfile(ChargingProfile chargingProfile) {
+  public boolean addChargingProfile(ChargingProfile chargingProfile) {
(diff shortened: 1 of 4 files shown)


pglob (changes requested):
> It is not possible to set a ChargingProfile with purpose set to TxProfile without presence of an active transaction, or in advance of a transaction.

If this happens, send a response rejection. `ChargingProfileHandler` has access to the transaction handler already, so this restriction can be checked there.

backend/src/main/java/com/sim_backend/electrical/ChargingProfileHandler.java
```diff
@@ -49,21 +49,30 @@ public ChargingProfileHandler(TransactionHandler transactionHandler, OCPPWebSock
...
       chargingTuples = chargingTuplesMax;
     } else if (purpose.equals(ChargingProfilePurpose.TX_DEFAULT_PROFILE)) {
       chargingTuples = chargingTuplesTxDefault;
-    } else {
+    }
+
+    // Case where purpose set to TxProfile
+    else {
+      // Not possible to set a ChargingProfile with this purpose without an active transaction.
+      if (!transactionHandler.getStartInProgress().get()) {
```
pglob, line 65: `getStartInProgress` is a flag that indicates we're in the middle of a start transaction flow. It goes to false after we receive a `StartTransactionResponse`.

When there's no active transaction, `transactionId` will be -1.

pglob (approved):
Just one more thing:

PR #116, opened 2025-03-16
Record Tx messages when sending instead of pushing

pglob:
Previously, messages would show in the "Sent Messages" log even when the simulator was offline.

### DIFF
--- a/backend/src/main/java/com/sim_backend/rest/controllers/MessageController.java
+++ b/backend/src/main/java/com/sim_backend/rest/controllers/MessageController.java
@@ -85,7 +85,7 @@ public class MessageController extends ControllerBase {
     Charger charger = getChargerID(ctx);
     if (charger == null) return;
     if (!checkWsClient(charger, ctx)) return;
-    Authorize msg = new Authorize();
+    Authorize msg = new Authorize(charger.getConfig().getIdTag());
 
     charger.getWsClient().pushMessage(msg);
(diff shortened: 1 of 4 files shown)


Xerasin (commented):
Unrelated to your changes but I noticed an issue while poking around that would be exacerbated by this. 

`[recordTxMessage](https://github.com/pglob/ocpp-charger-sim/blob/c73596805f99d490e45cb5afa3ced480b7f0848c/backend/src/main/java/com/sim_backend/websockets/OCPPWebSocketClient.java#L165)` makes an assumption that the last received request is always the one the response is for. (There is a `rxRequestName` String storing this every time we receive a request)

pglob:
@Xerasin is there a case where that assumption won't hold? I know that it will be incorrect if the central system breaks the protocol and sends multiple requests.

Xerasin:
Now that you mention it, not really except that case in which they break protocol. Just gives a bit of code stink, up to you then.

Xerasin (approved):

Xerasin:
Sorry about that

pglob:
@Xerasin all good lol. Changed it so rx request names are tracked by message ID. 

Also informally tested received many calls at the same time, nothing bad happened which is good

Xerasin (approved):
Looks good! Not at all needed but `rxRequestNames.put(msgId, messageName);` could be moved to line 428?

PR #121, opened 2025-03-20
Remote Transactions and Observer

JayQuack:
I made a new branch with changed files.
If it doesn't work, can you please help me to deal with a previous branch tomorrow?

### DIFF
--- a/backend/src/main/java/com/sim_backend/config/ConfigurationRegistry.java
+++ b/backend/src/main/java/com/sim_backend/config/ConfigurationRegistry.java
@@ -57,6 +57,9 @@ public class ConfigurationRegistry {
   /** A sample meter value interval. */
   private int MeterValueSampleInterval = 30;
 
+  /** Authorize data for Remote Transaction */
+  private boolean authorizeRemoteTxRequests;
+
   /** Sample meter value data. */
   private MeterValuesSampledData meterValuesSampledData =
       MeterValuesSampledData.ENERGY_ACTIVE_IMPORT_REGISTER;
@@ -64,6 +67,7 @@ public class ConfigurationRegistry {
   public ConfigurationRegistry(String idTag, String centralSystemUrl) {
     this.idTag = idTag;
     this.centralSystemUrl = centralSystemUrl;
+    this.authorizeRemoteTxRequests = true;
   }
 
   private static final String CONFIG_FILE_PATH = "config.properties";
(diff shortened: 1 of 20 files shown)


backend/src/main/java/com/sim_backend/websockets/observers/RemoteStopTransactionObserver.java
pglob, line 1: Just tested the remote stop with Ranger EV's backend, looks good 😁

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStartTransactionObserver.java
```diff
@@ -0,0 +1,97 @@
...
+                request.getConnectorId(),
+                request.getIdTag(),
+                transactionHandler.getTransactionId(),
+                transactionHandler.getElec(),
+                transactionHandler.getStartInProgress());
+        transactionHandler.setIdTag(request.getIdTag());
+      }
+    }
+
+    request.getChargingProfile().setChargingProfilePurpose(ChargingProfilePurpose.TX_PROFILE);
```
pglob, line 74: The charging profile may be null, so check that before using it.

backend/src/main/java/com/sim_backend/config/ConfigurationRegistry.java
```diff
@@ -57,13 +57,17 @@ public class ConfigurationRegistry {
   /** A sample meter value interval. */
   private int MeterValueSampleInterval = 30;
 
+  /** Authorize data for Remote Transaction */
+  private boolean authorizeRemoteTxRequests;
```
pglob, line 61: `authorizeRemoteTxRequests` needs to be added to the `GetConfigurationObserver` and the `ChangeConfigurationObserver`.

backend/src/main/java/com/sim_backend/websockets/observers/RemoteStartTransactionObserver.java
```diff
@@ -0,0 +1,97 @@
...
+        new RemoteStartTransactionResponse(request, "Accepted");
+    client.pushMessage(response);
+  }
+
+  // Check if it's okay to start transaction
+  public boolean checkState(ChargerState state, AtomicBoolean startinprogress) {
+    if (state != ChargerState.Available || startinprogress.get()) {
+      return false;
+    }
+    if (!startinprogress.compareAndSet(false, true)) {
```
pglob, line 91: This should be moved into the else block at line 61. If `startinprogress` gets set here, `transactionHandler.startCharging(request.getConnectorId(), request.getIdTag());` will always return because it thinks a start transaction is already in progress.

pglob (changes requested):
Just a few things left

pglob (approved):
Whoops, sorry for the delay.

